PR #5, opened 2020-09-03
updated storybook and lerna + created DetailComponent component

vysakh-menon-aot:
Issue #: /bcgov/entity#4812

*Description of changes:*
- updated Storybook to latest version
- updated folder structure per latest version
- created shared component "DetailComment"

*Notes:*

Created npm organization [https://www.npmjs.com/settings/bcrs-shared-components/packages](https://www.npmjs.com/settings/bcrs-shared-components/packages), all packages will be published here.

- Login to npm `npm login`
- To publish new version execute `lerna version` -> change version, this will update version in package.json and commit those changes.
- Run `lerna publish from-package` to explicitly publish packages where the latest version is not present in the registry. Yet to figure out `lerna publish from-git`, good for automation.

To consume the npm package execute `npm i @bcrs-shared-components/detail-comment` in other app's

Note: This app can still be refined before using it.

Examples of Storybook are contained in "src/stories" folder.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the bcrs-entities-create-ui license (Apache 2.0).


.storybook/preview.js
```diff
@@ -0,0 +1,18 @@
+import 'core-js/stable' // to polyfill ECMAScript features
+import 'regenerator-runtime/runtime' // to use transpiled generator functions
+import '@mdi/font/css/materialdesignicons.min.css' // ensure you are using css-loader
+import Vue from 'vue'
+import Vuetify from 'vuetify'
+import 'vuetify/dist/vuetify.min.css'
+import Vuelidate from 'vuelidate'
+import Affix from 'vue-affix'
+import Vue2Filters from 'vue2-filters' // needed by SbcFeeSummary
```
severinbeauvais, line 9: I think Affix and Vue2Filters are not needed. Please verify and remove accordingly.

lerna.json
```diff
@@ -1,6 +1,16 @@
 {
+  "version": "independent",
+  "npmClient": "npm",
   "packages": [
-    "packages/*"
+    "src/components/*"
```
severinbeauvais, line 5: 👍

lerna.json
```diff
@@ -1,6 +1,16 @@
...
   ],
-  "version": "0.0.0"
+  "command": {
+    "publish": {
+      "ignoreChanges": [
+        "*.stories.ts",
+        "*.md"
+      ],
+      "message": "chore(release): publish"
```
severinbeauvais, line 13: We might want to change this. Anyone have an opinion?

src/stories/Header.vue
```diff
@@ -0,0 +1,59 @@
+<template>
```
severinbeauvais, line 1: I don't think we want Header in here, do we? Was this a test?

src/stories/Page.vue
```diff
@@ -0,0 +1,90 @@
+<template>
```
severinbeauvais, line 1: What is this component for?

src/components/DetailComment/DetailComment.stories.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export default {
+  title: 'component/DetailComment',
+  component: DetailComment
+}
+
+export const Default = () => ({
+  components: { DetailComment },
+  template: '<detail-comment />'
+})
```
severinbeauvais, line 11: It would be nice to have some use cases in here.
This will be done in the future when we refactor this component (other ticket). FYI.

package.json
```diff
@@ -2,20 +2,73 @@
...
+    "style-loader": "^1.2.1",
+    "typescript": "^4.0.2",
+    "vue-class-component": "^7.2.2",
+    "vue-cli-plugin-vuetify": "^2.0.5",
+    "vue-debounce-decorator": "^1.0.1",
+    "vue-plugin-helper-decorator": "^0.0.12",
+    "vue-property-decorator": "^8.4.0",
+    "vue-template-compiler": "^2.6.11",
+    "vue-test-utils-helpers": "git+https://github.com/AmpleOrganics/vue-test-utils-helpers.git",
+    "vuex-module-decorators": "^0.16.1"
```
severinbeauvais, line 72: Well this is a huge list and it would be nice to prune it later (not now).

For one, I believe `flush-promises` is only used in the Jest unit tests.

Speaking of which... I think it's possible to run the unit tests in this project. Did you come across anything (that we can put in a new ticket for future investigation)?

severinbeauvais (commented):
Please see my comments.

.storybook/preview.js
vysakh-menon-aot: I copied these from filings-ui, since we are planning to move common components here. For DetailComment we may not need all of it

package.json
vysakh-menon-aot: As I commented in task this app still be refined before using it. I kept as is since its a POC

src/components/DetailComment/DetailComment.stories.ts
vysakh-menon-aot: examples are in src/stories folder

src/stories/Header.vue
vysakh-menon-aot: These are samples for the references and it won't be published as npm packages

.storybook/preview.js
severinbeauvais: We may or may not move sbc-common-components to this new library. The libraries have slightly different purposes.

If you have another commit, please remove these imports if not needed. If you don't have another commit, we can remove these imports later.

src/components/DetailComment/DetailComment.stories.ts
severinbeauvais: Is there an example for DetailComment?

src/stories/Header.vue
severinbeauvais: Fair enough. I see a lot of sample code in here we may be able to copy for our actual components.

src/components/DetailComment/DetailComment.stories.ts
vysakh-menon-aot: do you mean the available props and events?

severinbeauvais: Yes.

.storybook/preview.js
vysakh-menon-aot: removed

severinbeauvais:
Lekshmi mentioned to me that the dist folder in each component doesn't need to be checked in (since it's not published to npm, right?)

Also the xxx.stories.ts files don't serve any purpose in npm, so they can be ignored too.

Can you see about updating the git ignore file accordingly please?

src/components/DetailComment/DetailComment.stories.ts
vysakh-menon-aot: image

vysakh-menon-aot: you can see that in screen like this

vysakh-menon-aot:
> Lekshmi mentioned to me that the dist folder in each component doesn't need to be checked in (since it's not published to npm, right?)
> 
> Also the xxx.stories.ts files don't serve any purpose in npm, so they can be ignored too.
> 
> Can you see about updating the git ignore file accordingly please?

I tried npm ignore file to ignore stories.ts, unfortunately it didn't work. What we could do though is to keep stories.ts in separate folder like we do for test.

src/stories/Page.vue
severinbeauvais: It's a sample that we might want to copy from, same as Header.

src/components/DetailComment/DetailComment.stories.ts
severinbeauvais: OK for now.

lerna.json
severinbeauvais: Something like "Component version updated by Lerna" ?

vysakh-menon-aot: Version will not be updated here since I changed versioning independently for each component `"version": "independent"`. It will be in the package.json of each component

severinbeauvais:
> I tried npm ignore file to ignore stories.ts, unfortunately it didn't work. What we could do though is to keep stories.ts in separate folder like we do for test.

Would updating https://github.com/bcgov/bcrs-shared-components/blob/master/.gitignore work?

vysakh-menon-aot:
> > I tried npm ignore file to ignore stories.ts, unfortunately it didn't work. What we could do though is to keep stories.ts in separate folder like we do for test.
> 
> Would updating https://github.com/bcgov/bcrs-shared-components/blob/master/.gitignore work?

We need *.stories.ts in git, so i dont think its valid

.npmignore
```diff
@@ -0,0 +1 @@
+.stories.ts
```
severinbeauvais, line 1: What does this do?

severinbeauvais: fixed

severinbeauvais (approved):
Looks great! Thanks for the small updates.

Please prod the other reviewers (in RC) to give them a chance to see this. Give them a chance to comment, and then I'm OK if you merge this.

.npmignore
vysakh-menon-aot: I am trying it

lekshmimallika-aot (approved):

PR #6, opened 2020-09-16
4549 Insert staff payments into incorp app correction and make this a common component

vysakh-menon-aot:
Issue #: /bcgov/entity#4549

Description of changes:
 - StaffPayment component 
 - Setup testing 
 - Storybook config

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the lear license (Apache 2.0).


src/components/DetailComment/DetailComment.stories.ts
```diff
@@ -1,11 +1,19 @@
...
+  argTypes: {
+  }
 }
 
-export const Default = () => ({
+const Template = (args, { argTypes }) => ({
+  props: Object.keys(argTypes),
   components: { DetailComment },
-  template: '<detail-comment />'
+  template: '<detail-comment v-bind="$props" />'
```
severinbeauvais, line 13: This is cool but I'd like to see a comment that `$props` comes from `args` below.

src/components/StaffPayment/StaffPayment.stories.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const Template = (args, { argTypes }) => ({
+  props: Object.keys(argTypes),
+  components: { StaffPayment },
+  template: '<staff-payment v-bind="$props" />'
+})
+
+export const Default = Template.bind({})
+Default.args = {
+  staffPaymentData: {
```
severinbeauvais, line 19: Would it be good to specify the interface here? Eg,
`Default.args: StaffPaymentIF = { ... }`

src/components/StaffPayment/package.json
```diff
@@ -0,0 +1,12 @@
...
+    "name": "@bcrs-shared-components/staff-payment",
+    "version": "1.0.3",
+    "publishConfig": {
+        "access": "public"
+    },
+    "dependencies": {
+        "@bcrs-shared-components/enums": "^1.0.0",
+        "@bcrs-shared-components/interfaces": "^1.0.2",
+        "vue-property-decorator": "^8.4.0"
+    }
```
severinbeauvais, line 11: Why is Vuetify not needed here?

src/shims-tsx.d.ts
```diff
@@ -0,0 +1,13 @@
...
+  namespace JSX {
+    // tslint:disable no-empty-interface
+    interface Element extends VNode {}
+    // tslint:disable no-empty-interface
+    interface ElementClass extends Vue {}
+    interface IntrinsicElements {
+      [elem: string]: any
+    }
+  }
+}
```
severinbeauvais, line 13: Is this file needed? If not, please remove it.

Also for src/shims-vue.d.ts.

Also for src/utils/sleep.ts.

severinbeauvais (approved):
Awesome updates!

Nice cleanup of files to ignore! 👍 

Do you have steps to run the Jest tests?

I see you are committing a copy of the fonts. Is there an alternative? I know they're just used for Storybook here, but still it's another repo to update if they change. (This is not a big deal though.)

Looks good to me!

src/shims-tsx.d.ts
vysakh-menon-aot: Yes. shims-vue to avoid vs code lint error for .vue file and sleep is been used in unit test

src/components/StaffPayment/package.json
vysakh-menon-aot: then we have to mention all other dependencies including vue. do you think its needed?

src/components/StaffPayment/StaffPayment.stories.ts
vysakh-menon-aot: Will do the change

severinbeauvais: Cool.
It probably doesn't help much for Storybook, but it makes it explicit if a developer copies this code.

src/components/StaffPayment/package.json
severinbeauvais: Upon thinking about this, I believe you are right: this isn't needed.

Currently, in this project, Vue/Vuetify is declared by Storybook (in `preview.js`). This is similar to the app, where `main.ts` declares Vue/Vuetify. In both cases, the component doesn't need to know anything about this.

Therefore, here, I agree that the component doesn't need to import Vue or Vuetify -- it should just expect Storybook/app to provide it.

PR #9, opened 2020-10-08
Certify Component

rarmitag:
Certify Component with storybook.

Need to pass currentDate property to component instead of reading it from the store


src/components/Certify/Certify.vue
```diff
@@ -0,0 +1,157 @@
+<template>
+  <v-card flat id="AR-step-4-container">
```
cameron-freshworks, line 2: We could probably rename the ID to something more generic at this point. Something as simple as `certify-container`

src/components/Certify/Certify.vue
```diff
@@ -0,0 +1,157 @@
...
+  </v-card>
+</template>
+
+<script lang="ts">
+import { Component, Vue, Prop, Emit } from 'vue-property-decorator'
+import { mapState } from 'vuex'
+
+@Component({
+  computed: {
+  // ...mapState(['currentDate'])
```
cameron-freshworks, line 46: Could probably remove the comment.

cameron-freshworks (approved):
Looks great! Just a few minor question/comments

vysakh-menon-aot (changes requested):

severinbeauvais:
@rarmitag There should only be 1 Assignee -- you.

src/components/Certify/package.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "@bcrs-shared-components/certify",
+    "version": "1.0.3",
```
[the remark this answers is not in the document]
severinbeauvais: 0.0.0 or 1.0.0 ?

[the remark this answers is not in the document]
vysakh-menon-aot: When we run lerna version it modify the package with 1.0.0 if we set 0.0.0 before. If we set 1.0.0 before that then it will prompt to upgrade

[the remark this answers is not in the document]
severinbeauvais: Summary: please set initial version to "0.0.0".

tests/unit/Certify.spec.ts
```diff
@@ -0,0 +1,175 @@
+//
+// Copyright © 2019 Province of British Columbia
```
severinbeauvais, line 2: If we're putting in the copyright notice (which we are not doing consistently atm, so this could go either way) then please set this to 2020 :)

src/components/Certify/Certify.vue
```diff
@@ -0,0 +1,157 @@
...
+    <div class="certify-container">
+      <div class="certifiedby-container">
+        <label>Legal Name</label>
+        <div class="value certifiedby">
+          <v-text-field
+            filled
+            persistent-hint
+            id="certified-by-textfield"
+            label="Person's legal name"
+            hint="Legal name of authorized person"
```
severinbeauvais, line 12: This label/hint/rule was taken from business-filings-ui but they are different in bcrs-business-create-ui...

What should these be in business-edit-ui? Let's change these to the requirements. (We can worry about passing more props to set this per project in a future version of this component.)

src/components/Certify/Certify.vue
```diff
@@ -0,0 +1,157 @@
...
+            :rules="[(v) => !!v || 'A person\'s legal name is required.']"
+            @input="emitCertifiedBy($event)"
+          />
+        </div>
+      </div>
+      <v-checkbox
+        :value="isCertified"
+        @change="emitIsCertified($event)"
+        id="isCertified-checkbox"
+      >
```
severinbeauvais, line 23: As per bcrs-business-create-ui, you should probably add "hide-details" prop here.

https://github.com/bcgov/bcrs-business-create-ui/blob/928c46e881008572ac60893ce7a3eec34d48c210/src/components/ReviewConfirm/Certify.vue#L20

src/components/Certify/Certify.vue
```diff
@@ -0,0 +1,157 @@
...
+        <template slot="label">
+          <div class="certify-stmt">
+            I,
+            <strong>{{ trimmedCertifiedBy || "[Legal Name]" }}</strong>
+            , certify that I have relevant knowledge of the
+            {{ entityDisplay || "association" }} and that I am authorized to
+            make this filing.
```
severinbeauvais, line 30: Again here, please check the required wording for business-edit-ui.

src/components/Certify/Certify.vue
```diff
@@ -0,0 +1,157 @@
...
+          </div>
+        </template>
+      </v-checkbox>
+      <p class="certify-clause">Date: {{ currentDate }}</p>
+      <p class="certify-clause">{{ message }}</p>
```
severinbeauvais, line 35: Does business-edit-ui have a requirement to display email addresses here?

https://github.com/bcgov/bcrs-business-create-ui/blob/928c46e881008572ac60893ce7a3eec34d48c210/src/components/ReviewConfirm/Certify.vue#L39

src/components/Certify/Certify.vue
severinbeauvais: ~~I agree with Cam. Let's change this id to `certify-container` (similar to StaffPayment shared component).~~

Doh, id is already in use.

src/components/Certify/Certify.vue
```diff
@@ -0,0 +1,157 @@
...
+}
+</script>
+
+<style lang="scss" scoped>
+// @import '@/assets/styles/theme.scss';
+
+#AR-step-4-container {
+  margin-top: 1rem;
+  padding-bottom: 0.5rem;
+  padding-top: 1rem;
```
severinbeauvais, line 110: Similar to other components (such as StaffPayment), the component itself should not specify the whitespace around itself -- it should be as tight as possible -- then its parent can set the appropriate whitespace as needed.
~So please remove margin and padding from here.~

src/components/Certify/Certify.vue
severinbeauvais: Never mind. Just leave this alone. It's not ideal but it's consistent with shared StaffPayment component.

src/components/Certify/Certify.vue
severinbeauvais: The idea is sound but this is consistent with shared StaffPayment component.
Just leave this alone for now.

severinbeauvais (changes requested):

severinbeauvais:
What is the Zenhub ticket for this PR?

Does this ticket also require integrating this component into business-edit-ui?

severinbeauvais:
> @rarmitag There should only be 1 Assignee -- you.

I just updated this.

src/components/Certify/Certify.vue
severinbeauvais: This is still not in line with the design. Please see https://preview.uxpin.com/c53183da94b200338185d9557b11dd9a860f9092#/pages/131426862/simulate/no-panels

src/components/Certify/Certify.vue
```diff
@@ -0,0 +1,155 @@
...
+    <div class="certify-container">
+      <div class="certifiedby-container">
+        <label>Legal Name</label>
+        <div class="value certifiedby">
+          <v-text-field
+            filled
+            persistent-hint
+            id="certified-by-textfield"
+            label="Person's legal name"
+            hint="Enter the legal name of the person authorized to complete and submit these changes."
```
severinbeauvais, line 12: I know you just changed this, but now that I'm looking at the design it doesn't match.
https://preview.uxpin.com/c53183da94b200338185d9557b11dd9a860f9092#/pages/131426862/simulate/no-panels

Do you have a later design that asks for the wording above?

severinbeauvais (approved):
Check some of the verbiage against the design. When it's OK you can merge this.

vysakh-menon-aot (approved):

PR #15, opened 2021-02-01
5167 fee summary component

cameron-freshworks:
*Issue #:* /bcgov/entity#5167

*Description of changes:*

* Added Sass as file type to load.

* Updated some dependencies.

* Added FeeSummary composite component.

* Updated enums/interfaces.


vysakh-menon-aot (approved):

src/assets/styles/theme.scss
```diff
@@ -57,4 +57,10 @@ $BCgovInputBG: #ffffff;
...
 
 // Icon colors
-$BCgovIconBlue: #2196F3
+$BCgovIconBlue: #2196F3;
+
+  // App Colors
+$app-blue: #1669bb; // same as the Vuetify theme primary
+$app-dk-blue: #38598a;
+$app-red: #d3272c;
+$app-green: #1a9031,
```
severinbeauvais, line 66: With these common colors, enums (below), etc, now we really need to make sure our framework is the same between UI projects, to prevent conflicts.

No action needed here, just a mental note to use similar files/variables whenever possible.

severinbeauvais: Or maybe we should move our theme, enums, etc to THIS project and just import them???

severinbeauvais (commented):
Very good!

src/assets/styles/theme.scss
cameron-freshworks: Absolutely, I was thinking the same thing, just not sure how best to do this.
Maybe we should have mini UI meeting and brainstorm a way to best approach this!

severinbeauvais: Sure, set up a meeting. See if you can get @jeznorth and/or @saravankumarpa and/or @sumesh-aot to come, as we all have a vested interest in this. Thx.

PR #16, opened 2021-02-04
added Staff Comments component

severinbeauvais:
*Issue #:* /bcgov/entity#6257

*Description of changes:*
- updated some styles
- added Staff Comment component
- updated placeholder color
- added Comment IF
- updated Form IF
- added minimal Date Mixin
- added minimal Filing Mixin
- added workflow to README
- updated axios package
- added global decorator to correctly initialize Vuetify
- updated Storybook templates
- deleted unneeded mixins package.json
- added axios mocking!
- implemented Staff Comment unit tests
- updated Storybook packages


.storybook/preview.js
```diff
@@ -1,20 +1,29 @@
...
 import '@/assets/styles/overrides.scss'
 
-Vue.use(Vuetify, {
-  iconfont: 'mdi'
-})
+Vue.use(Vuetify)
 Vue.use(Vuelidate)
 
+addDecorator(() => ({
+  vuetify: new Vuetify({ iconfont: 'mdi' }),
```
severinbeauvais, line 19: This little line here was required to "configure" Vuetify to work correctly. Before this I was getting an error, "Cannot read property 'dark' of undefined at VueComponent.appIsDark".

docs/iframe.html
```diff
@@ -4,6 +4,41 @@
     display: none;
   }
 
+  .sb-show-main.sb-main-centered {
```
severinbeauvais, line 7: Please ignore this file and all others in /docs. These are the Storybook pages.

src/interfaces/form-interface.ts
```diff
@@ -2,5 +2,6 @@ import Vue from 'vue'
 
 export interface FormIF extends Vue {
   reset(): void;
+  resetValidation(): void;
```
severinbeauvais, line 5: Ref: https://vuetifyjs.com/en/api/v-form/#functions

(Also, my new component uses this.)

src/components/StaffComments/StaffComments.vue
```diff
@@ -0,0 +1,252 @@
+<template>
```
severinbeauvais, line 1: This is exact same component that you already reviewed in Staff Comments Part 1 PR.

Except... I had to remove conditional chaining (`a?.b`) as that caused a JS error. This is likely due to the TS version used but I can find nothing out of date.

This can be fixed in the future.

src/components/StaffComments/StaffComments.stories.ts
```diff
@@ -0,0 +1,64 @@
...
+    comment: {
+      comment: "First comment.",
+      submitterDisplayName: "Tester",
+      timestamp: "2019-06-02T19:22:59.003777+00:00"
+    }
+  }
+]
+
+// set the mock adapter on the default instance
+const mock = new MockAdapter(axios);
```
severinbeauvais, line 22: This is cool. Check it out.

src/assets/styles/overrides.scss
```diff
@@ -67,3 +67,16 @@
 .v-application--is-ltr .v-list-item__icon:first-child {
   margin-right: 1rem;
 }
+
+// Error Colours
```
severinbeauvais, line 71: PS I don't think Storyboook (React?) is correctly importing our overrides stylesheet. I noticed one instance where our v-btn's styles were not being used.

(This is a future problem to fix.)

lerna.json
```diff
@@ -3,8 +3,8 @@
   "npmClient": "npm",
   "packages": [
     "src/components/*",
-    "src/interfaces",
-    "src/enums"
+    "src/enums",
+    "src/interfaces"
   ],
```
severinbeauvais, line 8: I did not add the new mixins folder to this (and it seems to work fine). As explained in another comment, my new component uses `@mixins` which grabs the local mixins folder (either here in bcrs-shared-components or when imported in our main web apps).

src/components/StaffComments/StaffComments.vue
```diff
@@ -0,0 +1,256 @@
...
+        </div>
+      </v-card-text>
+    </v-card>
+  </v-menu>
+</template>
+
+<script lang="ts">
+import { Component, Mixins, Prop } from 'vue-property-decorator'
+import { CommentIF, FormIF } from '@/interfaces'
+import { DateMixin, FilingMixin } from '@/mixins'
```
severinbeauvais, line 79: Note that this new component uses the "local" interfaces and mixins.

src/components/StaffComments/StaffComments.vue
```diff
@@ -0,0 +1,256 @@
...
+
+  /** The new comment's text. */
+  private comment: string = null
+
+  /** Whether a comment is being saved. */
+  private isSaving = false
+
+  /** The number of chars remaining in the new comment. */
+  private get charsRemaining (): number {
+    const length = this.comment ? this.comment.length : 0 // comment may be null
```
severinbeauvais, line 113: I had to remove the previous conditional chaining (`a?.b`) as that caused a JS error. This is likely due to the TS version used but I can find nothing out of date.

(This is a future problem to fix.)

vysakh-menon-aot (approved):

lerna.json
severinbeauvais: I might be a good idea to publish/version the mixins folder, so as to keep changes managed. However, I don't expect UIs to import these mixins (although they could).

src/components/StaffComments/StaffComments.vue
```diff
@@ -0,0 +1,256 @@
...
+    await this.fetchStaffComments()
+  }
+
+  /** Fetches the staff comments from the API. */
+  private async fetchStaffComments (): Promise<void> {
+    const url = `businesses/${this.businessId}/comments`
+    this.comments = await this.axios.get(url)
+      .then(res => this.flattenAndSortComments(res && res.data && res.data.comments))
+      .catch(() => [])
+  }
```
cameron-freshworks, line 144: Were you able to make any mocked requests in Storybook?
No action required here, just curious because i'd like to know how :)

cameron-freshworks (approved):
Looks great, nice work on the updated readme. thanks!

src/components/StaffComments/StaffComments.vue
severinbeauvais: Yes!

cameron-freshworks: Very cool, i wonder if my troubles were due too using the `SbcFeeSummary` common component imported into the shared component, where the actual requests were from that import.. 🤔 

Either way, well done!

PR #18, opened 2021-02-16
added Corp Type Module

severinbeauvais:
*Issue #:* /bcgov/entity#6393

**NB: I have tested this in-situ in Filings UI (and will be importing it there shortly).**

*Description of changes:*
- added Corp Type Module
- added unit test file

### DIFF
--- a/src/modules/corp-type-module/corp-type-module.ts
+++ b/src/modules/corp-type-module/corp-type-module.ts
@@ -1,0 +1,396 @@
+//
+// Module containing corp type enums, interface and access functions.
+//
+
+/**
+ * Corp type codes, used as keys to get corp info.
+ * (sorted by code)
+ */
+export enum CorpTypeCd {
+  EXTRA_PRO_A = 'A',
+  EXTRA_PRO_B = 'B',
+  BC_COMPANY = 'BC',
+  BENEFIT_COMPANY = 'BEN',
+  CONTINUE_IN = 'C',
+  BC_CCC = 'CC',
+  CCC_CONTINUE_IN = 'CCC',
+  CEMETARY = 'CEM',
+  COOP = 'CP',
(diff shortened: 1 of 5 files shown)


src/modules/corp-type-module/corp-type-module.ts
```diff
@@ -0,0 +1,396 @@
...
+  CONT_IN_SOCIETY = 'CS',
+  ULC_CONTINUE_IN = 'CUL',
+  EXTRA_PRO_REG = 'EPR',
+  FINANCIAL = 'FI',
+  FOREIGN = 'FOR',
+  PARTNERSHIP = 'GP',
+  LIBRARY = 'LIB',
+  LICENSED = 'LIC',
+  LL_PARTNERSHIP = 'LL',
+  LIMITED_CO = 'LLC',
```
cameron-freshworks, line 29: Just curious what these are mapped from. 
In Lear a limited company is 'BC'. 
Also, in Name Request an 'LLC' is a Limited `Liability` Company'

Seems challenging to keep things in sync! Are `corpType`'s mapping back to something in Auth?

cameron-freshworks: Maybe i just took the abbreviated name too literally, as it is call `LIMITED_CO`. So maybe i'm mistaken here..

cameron-freshworks (approved):
Looks good, just a minor question!

src/modules/corp-type-module/corp-type-module.ts
severinbeauvais: This is from a spreadsheet that Linda attached to the ticket.

For these key names I used the "short desc" as that is sort-of meaningful.

There is another list, "entity types", which is used by NRs, which contains some other types. Combining these will require another ticket... and maybe getting the BAs to take a crack at it first.

Specifically, here, don't worry about "LMITED_CO = LLC". The first is an internal key name and the second is the actual corp type cd in the dbs.

PR #19, opened 2021-02-18
6448 Shared Contact Component

cameron-freshworks:
*Issue #:* /bcgov/entity#6448

Visual Design for reference: https://projects.invisionapp.com/share/6EYT1CDGYVN#/screens/444755344

*Description of changes:*

* Adapted the Contact info to be a shared component.

* Existing Logic and Validation refactors.

* Additional Unit tests.


src/components/ContactInfo/ContactInfo.vue
```diff
@@ -0,0 +1,365 @@
...
+          There is no fee to change Registered Office Contact Information. Any changes made will be applied immediately.
+        </v-flex>
+      </v-layout>
+
+      <v-form
+        v-model="formValid"
+        ref="editContactForm"
+        name="business-contact-form"
+        class="business-contact-form pt-5"
+        @submit.prevent="submitContact()"
```
severinbeauvais, line 111: What exactly does this do?

severinbeauvais: Oh, does it catch the submit event from the Done button?

src/components/ContactInfo/ContactInfo.vue
```diff
@@ -0,0 +1,365 @@
...
+      (v: string) => !!v || 'Confirm email address',
+      (v: string) => v.toString() === this.contactInfo.email || 'Email addresses must match'
+    ]
+    this.phoneRules = [
+      (v: any) => !v || (v.length === 0 || v.length === 14) || 'Phone number is invalid'
+    ]
+
+    // Await the applied rules and validate form
+    await Vue.nextTick()
+    this.$refs.editContactForm.validate()
```
severinbeauvais, line 304: This is an interesting way to validate only after the user clicks Save. Does it work well?

Have you ever gotten lazy validation to work? In theory that's an alternative.

src/components/ContactInfo/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@bcrs-shared-components/contact-info",
+  "version": "1.0.0",
```
severinbeauvais, line 3: I recommend "0.0.0" so that when Lerna first versions it, then you can select "1.0.0".

src/interfaces/contact-interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface ContactIF{
+  email: string
+  confirmEmail?: string
```
severinbeauvais, line 3: Isn't this a required field?

src/interfaces/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@bcrs-shared-components/interfaces",
-    "version": "1.0.9",
+    "version": "1.1.0",
```
severinbeauvais, line 3: You don't do this. Lerna does.

severinbeauvais (approved):
The code looks really good. Clean, good use of class names, easy to review :)

Just a few questions/ possible tweaks but gtg when you're satisfied.

src/components/ContactInfo/ContactInfo.vue
cameron-freshworks: Absolutely!
Catches the default submit behaviour and calls the method we define instead.

src/components/ContactInfo/ContactInfo.vue
cameron-freshworks: This works pretty well, yea! 
By handling the validations and applying the rules on submit it prevents those validations triggers while a user is typing/editing and so on. 
Very often the feedback from UX is related to way Vue validator `rules` work, so this is just another approach. 
Name request also has another approach but it seemed very verbose, so just tried to keep it simple here and only apply the rules once a user was done with the form. 
Is lazy validation a feature of Vue or is it a alternative package perhaps?

src/components/ContactInfo/package.json
cameron-freshworks: oh nice, ok will do!

src/interfaces/contact-interface.ts
cameron-freshworks: I'm not sure to be honest. The field itself isn't a value that we save/store, so when we retrieve contact info, it isn't present.
The value is only required fpr submission events for validating, so it's sort of in-between, but my thoughts were it wasn't required in the model, so perhaps it should be optional.

src/interfaces/package.json
cameron-freshworks: oh nice, will correct, thanks!

src/components/ContactInfo/ContactInfo.vue
severinbeauvais: https://vuejs.org/v2/guide/forms.html#lazy

src/interfaces/contact-interface.ts
severinbeauvais: OK, agreed. I thought we submitted to the BE.

src/interfaces/package.json
severinbeauvais: If these steps needs tweaking, feel free to update them!
https://github.com/bcgov/bcrs-shared-components/blob/master/README.md

PR #20, opened 2021-02-21
6448 contact info patch

cameron-freshworks:
*NOTE:* - Looking at options for adapting for Corrections. It's pretty much ready, but waiting on design feedback. 
* Confirmed question with design team. g2g.

* Display the correct prop values when loading component in display mode.
* Fixed some validation logic to check the correct values before submitting.

### DIFF
--- a/src/components/ContactInfo/ContactInfo.vue
+++ b/src/components/ContactInfo/ContactInfo.vue
@@ -294,7 +294,8 @@ export default class ContactInfo extends Vue {
     ]
     this.confirmEmailRules = [
       (v: string) => !!v || 'Confirm email address',
-      (v: string) => !v || (v.toString() === this.$refs.editContactForm?.$el[0]?.value) || 'Email addresses must match'
+      (v: string) => !v || (v.toString() === (this.$refs.editContactForm && this.$refs.editContactForm.$el[0].value)) ||
+        'Email addresses must match'
     ]
     this.phoneRules = [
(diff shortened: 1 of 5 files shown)


src/components/ContactInfo/ContactInfo.vue
```diff
@@ -293,7 +294,8 @@ export default class ContactInfo extends Vue {
     ]
     this.confirmEmailRules = [
       (v: string) => !!v || 'Confirm email address',
-      (v: string) => v.toString() === this.contactInfo.email || 'Email addresses must match'
+      (v: string) => !v || (v.toString() === (this.$refs.editContactForm && this.$refs.editContactForm.$el[0].value)) ||
```
cameron-freshworks, line 297: I worked at trying to get our storybook to recognize optional chaining but was unsuccessful, so i opted to short circuit here. 

Will continue to do a little research with regards to storybooks issue with the TS configuration (i think that's what it is at least)

severinbeauvais: I, too, found that optional chaining didn't work in the bscs-shared-components project.

Maybe there's something missing from tsconfig.json ?

Good luck (may it work for all of us)!

severinbeauvais (approved):

src/components/ContactInfo/ContactInfo.vue
cameron-freshworks: haha thanks! 

Yes, i believe you are right. Will take a little bit of research but will be nice to figure it out!

PR #21, opened 2021-02-22
Contact version bump

cameron-freshworks:
*Note:* 
* I'm not really sure how these versions got out of sync, i followed the version update and publish sequence.. but for some reason, master didn't update these previously. 

* To fix missed version bump. 

* This will result in another patched version of these components without change beyond an updated interface version.

### DIFF
--- a/src/components/ContactInfo/package.json
+++ b/src/components/ContactInfo/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bcrs-shared-components/contact-info",
-  "version": "0.0.0",
+  "version": "1.0.0",
   "publishConfig": {
     "access": "public"
   },
--- a/src/components/ContactInfo/package.json
+++ b/src/components/ContactInfo/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bcrs-shared-components/contact-info",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "publishConfig": {
     "access": "public"
   },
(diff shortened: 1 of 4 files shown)


severinbeauvais (changes requested):
This should not happen if you follow the instructions in the README. (I followed those twice and it worked both times.)

Please try again, perhaps on a new branch.

Perhaps you need to rebase to bcgov/master before running Lerna?

cameron-freshworks:
> This should not happen if you follow the instructions in the README. (I followed those twice and it worked both times.)
> 
> Please try again, perhaps on a new branch.
> 
> Perhaps you need to rebase to bcgov/master before running Lerna?

cameron-freshworks:
> This should not happen if you follow the instructions in the README. (I followed those twice and it worked both times.)
> 
> Please try again, perhaps on a new branch.
> 
> Perhaps you need to rebase to bcgov/master before running Lerna?

Hmm this was a new branch after a Rebase.
Sort of stuck since Lerna thinks it has already publish 1.0.0 (which it has) but the version of the Contact is still 1.0.0 after the changes.

severinbeauvais:
> Sort of stuck since Lerna thinks it has already publish 1.0.0 (which it has) but the version of the Contact is still 1.0.0 after the changes.

According to this, Lerna has not updated the ContactInfo version:
https://github.com/bcgov/bcrs-shared-components/blob/master/src/components/ContactInfo/package.json

PR #22, opened 2021-02-23
Contact if dependency

cameron-freshworks:
*Issue #:* /bcgov/entity#6448

*Description of changes:*

* Updated the Interface dependency

* Rebuilt StoryBook


src/components/FeeSummary/package.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "@bcrs-shared-components/fee-summary",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "publishConfig": {
     "access": "public"
   },
   "dependencies": {
     "@bcrs-shared-components/enums": "^1.0.6",
-    "@bcrs-shared-components/interfaces": "^1.0.9",
+    "@bcrs-shared-components/interfaces": "^1.0.10",
```
severinbeauvais, line 9: Did you update the interface dependency here?

If not, I wonder how we would tell this component to use a fixed interface version instead -- perhaps remove the ^ ?

severinbeauvais (approved):

src/components/FeeSummary/package.json
severinbeauvais: Did interfaces change this time around? Weird...

cameron-freshworks: Weird indeed, i just added the dependency in the contact info. 
It appears that masters version was behind patch version, creating a diff? 
I really hope this sync's everything and is the last of these weird diffs 🙏

PR #24, opened 2021-02-25
added a bunch of interfaces

severinbeauvais:
*Issue #:* /bcgov/entity#5110

*Description of changes:*
- added ActionTypes enum
- added AlterationIF
- added BusinessIF
- added ContactPointIF
- added NameRequest interfaces
- added NameTranslationIF
- added ShareStructure interfaces
- updated README file
- added missing alias
- removed optional chaining
- rebuilt Storybook


.storybook/main.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
     // Alternately, for an alias:
     config.resolve.alias['@bcrs-shared-components/enums'] = path.resolve(__dirname, '../') + '/src/enums';
     config.resolve.alias['@bcrs-shared-components/interfaces'] = path.resolve(__dirname, '../') + '/src/interfaces';
+    config.resolve.alias['@bcrs-shared-components/corp-type-module'] = path.resolve(__dirname, '../') + '/src/modules/corp-type-module';
```
severinbeauvais, line 29: I had to make this specific because I originally didn't package "all modules" together -- just this one module.
Storybook needs this alias BTW.

src/enums/index.ts
```diff
@@ -1,3 +1,6 @@
+export * from './actionTypes'
 export * from './entityTypes'
 export * from './filingCodes'
 export * from './staffPaymentsOptions'
+
+export { CorpTypeCd, CorpClass } from '@bcrs-shared-components/corp-type-module'
```
severinbeauvais, line 6: Importing from myself, whoo.

src/enums/package.json
```diff
@@ -3,5 +3,8 @@
     "version": "1.0.6",
     "publishConfig": {
         "access": "public"
+    },
+    "dependencies": {
+        "@bcrs-shared-components/corp-type-module": "^1.0.0"
```
severinbeauvais, line 8: Enums now depends on this.

src/interfaces/alteration-interface.ts
```diff
@@ -0,0 +1,17 @@
...
+ * https://github.com/bcgov/business-schemas/blob/master/src/registry_schemas/schemas/alteration.json
+ */
+export interface AlterationIF {
+  provisionsRemoved: boolean
+  business: BusinessIF
+  nameRequest?: NameRequestIF
+  nameTranslations: NameTranslationIF[]
+  shareStructure: ShareStructureIF
+  contactPoint: ContactPointIF
+  [propName: string]: any // excess properties
```
severinbeauvais, line 16: This is require to "catch" any extra properties on the IF without causing a TS error.

For example, `alteration.severin` would be OK.

Ideally the IF would contain all and only the correct properties, but this allows extra properties to be added as needed without causing errors. That said, we _should_ go back and add legit properties to the IFs as we go along.

src/interfaces/business-interface.ts
```diff
@@ -0,0 +1,20 @@
+import { CorpTypeCd } from '@/enums'
+
+/**
+ * A filing's business object from the API. See:
+ * https://github.com/bcgov/business-schemas/blob/master/src/registry_schemas/schemas/business.json
+ */
+export interface BusinessIF {
```
severinbeauvais, line 7: Hey, I named this according to the schema (different than Edit UI).

src/interfaces/businessContact-interface.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * See:
+ * https://github.com/bcgov/business-schemas/blob/master/src/registry_schemas/schemas/contact_point.json
+ */
+export interface ContactPointIF{
```
severinbeauvais, line 5: This is also named according to the schema (different in Edit UI).

src/interfaces/nameRequest-interface.ts
```diff
@@ -0,0 +1,39 @@
...
+  addressLine1: string
+  addressLine2: string
+  addressLine3: string
+  city: string
+  countryTypeCode: string
+  postalCode: string
+  stateProvinceCode: string
+}
+
+export interface NameRequestIF {
```
severinbeauvais, line 33: The bcrs-shared-components compiler insisted on having this last (or rather, what it depends on above it).

src/interfaces/shareStructure-interface.ts
```diff
@@ -0,0 +1,27 @@
...
+  type?: string // Indicates whether class or series
+  name: string
+  priority: number
+  hasMaximumShares?: boolean
+  maxNumberOfShares: number
+  hasParValue?: boolean
+  parValue?: number
+  currency?: string
+  hasRightsOrRestrictions: boolean
+  series?: ShareClassIF[]
```
severinbeauvais, line 18: This is **not** the same as the schema. But it should be.

src/modules/corp-type-module/corp-type-module.ts
```diff
@@ -383,7 +383,8 @@ export function GetCorpInfoObject (cd: CorpTypeCd): CorpInfoIF {
  * @returns the description (or '' if not found)
  */
 export function GetCorpFullDescription (cd: CorpTypeCd): string {
-  return CorpInfoArray.find(obj => (cd === obj.corpTypeCd))?.fullDesc || ''
+  const item = CorpInfoArray.find(obj => (cd === obj.corpTypeCd))
+  return (item && item.fullDesc) || ''
```
severinbeauvais, line 387: Strangely, the optional chaining worked previously, but not once this module was imported into enums. Lots of Storybook errors. 👻

src/interfaces/alteration-interface.ts
cameron-freshworks: I like it!

src/interfaces/businessContact-interface.ts
cameron-freshworks: We should probably remove `ContactIF` and refer to this Schema as well, where used.

cameron-freshworks: I think just in the Contact component.

cameron-freshworks (approved):
Looks good!

src/interfaces/businessContact-interface.ts
severinbeauvais: It's used in the Alteration IF.

cameron-freshworks: ah ok. 
I believe it is the same IF, so we should delete the ContactIF and refer to this one perhaps.

severinbeauvais: Yes. That's what drove me to do this -- I didn't want to add a bunch of duplicate IFs in Filings UI. So now I will just import the common/shared ones.

PR #26, opened 2021-02-25
Updated IF imports to prevent TS import errors

cameron-freshworks:
* Updated imports to prevent TS build errors

* removed obsolete interface EntityTypes


src/interfaces/alteration-interface.ts
```diff
@@ -1,6 +1,6 @@
 import {
   BusinessIF, NameRequestIF, NameTranslationIF, ShareStructureIF, ContactPointIF
-} from '@/interfaces'
+} from './'
```
severinbeauvais, line 3: No, should probably point to
`@bcrs-shared-components/interfaces`.

severinbeauvais: Otherwise THIS file assumes the other files are in the local folder.

cameron-freshworks: that was my initial thought/attempt as well, though it threw TS errors in the Shared Components. Will have a closer look.

cameron-freshworks: Ah, ok it was throwing TS errors: 'cannot find module @bcrs-shared-components/interfaces' because..

This directory has it's own node_modules, where it is looking for this..

Screen Shot 2021-02-25 at 9 38 24 AM

So, the only option i see is to add the dependency (on itself!?) in the package.json

Screen Shot 2021-02-25 at 9 41 12 AM

cameron-freshworks: 😆  what do you think we should do?

cameron-freshworks: If you think this is ok, i will commit this.

severinbeauvais: Yeah that doesn't feel right.
Maybe this was correct after all: `import { blah} from './'`.

severinbeauvais: It means that project that import these interfaces will need to import all dependent interfaces too (eg, in index.ts file). I _think_ that's OK, not 100% sure though.

cameron-freshworks: kk, copy that! This would be g2g then

src/interfaces/filingData-interface.ts
```diff
@@ -1,10 +1,10 @@
-import { FilingCodes, EntityTypes } from '@/enums'
+import { CorpTypeCd, FilingCodes } from '@bcrs-shared-components/enums'
 
 /** Filing data object passed to the SBC Fee Summary. */
 export interface FilingDataIF {
   filingDescription?: string
   filingTypeCode: FilingCodes
-  entityType: EntityTypes
+  entityType: CorpTypeCd
```
severinbeauvais, line 7: Perfect.

severinbeauvais (approved):

PR #28, opened 2021-03-05
Shared ShareStructure, EditShareStructure, Confirm Dialog & ActionChip Components

cameron-freshworks:
Issue #: /bcgov/entity#5104

Description of changes:

Implemented Share Structure component as SHARED component.
Updated logic / validations / styling / Unit Testing.
Implemented Edit Share Structure component. ** NOT SHARED YET**

Implemented Confirm Dialog component as SHARED component.

Implemented ActionChip component as SHARED component.

Implemented required interfaces.

Fixed fee summary import.

Updated BC Benefit Company fullDesc


severinbeauvais:
> Also, if reviewers think it valuable to break out the `EditShareStructure` component as SHARED, please say so and i will do that as well 👍 

I don't think so. It's never used independently, right?

src/components/ActionChip/ActionChip.vue
```diff
@@ -0,0 +1,29 @@
...
+            x-small label color="primary" text-color="white">{{editedLabel(isCorrection)}}</v-chip>
+    <v-chip v-if="actionableItem.action === ActionTypes.REMOVED"
+            x-small label color="#E0E0E0" text-color="#212529">REMOVED</v-chip>
+  </div>
+</template>
+
+<script lang="ts">
+import { Component, Mixins, Prop } from 'vue-property-decorator'
+import { ActionTypes } from '@/enums'
+import { ActionableItemIF } from '@/interfaces'
```
severinbeauvais, line 15: Should this be `@bcrs-shared-components/interfaces` ? (and also for enums)

So that, in the app, it gets the shared interfaces?

Alternately, the app could re-export the shared interfaces. I did this once to prevent having to change a ton of stuff in the app but I don't think it's a good pattern. I think it's better if THIS component imports the things it needs directly.

Thoughts?

src/components/ConfirmDialog/ConfirmDialog.stories.ts
```diff
@@ -0,0 +1,37 @@
...
+const Template = (args, { argTypes }) => ({
+  vuetify: new Vuetify({ iconfont: 'mdi' }),
+  props: Object.keys(argTypes),
+  components: { ConfirmDialog },
+  template: '<confirm-dialog v-bind="$props" @hook:mounted="externalMount" />',
+  methods: {
+    async externalMount () {
+      const vm = this.$children[0] // target the component confirm-dialog
+      await Vue.nextTick() // wait that mounted() finished
+      vm.open(
```
severinbeauvais, line 21: Oh wow!

src/components/ShareStructure/ShareStructure.vue
```diff
@@ -0,0 +1,1058 @@
...
+      </template>
+    </v-data-table>
+  </v-card>
+</template>
+
+<script lang="ts">
+// Libraries
+import { Component, Emit, Mixins, Prop, Vue, Watch } from 'vue-property-decorator'
+import 'array.prototype.move'
+import { cloneDeep, isEqual, omit } from 'lodash'
```
severinbeauvais, line 399: Should add dependency on lodash to package.json.

src/components/ShareStructure/EditShareStructure.vue
```diff
@@ -0,0 +1,571 @@
...
+
+// Components
+import { ConfirmDialog } from '@/components/ConfirmDialog'
+
+// Interfaces and Enums
+import { ConfirmDialogType, FormIF, ShareClassIF } from '@/interfaces'
+
+// Mixins
+import { CurrencyLookupMixin } from '@/mixins'
+import { ActionTypes } from '@/enums'
```
severinbeauvais, line 177: As above, I think this should import from the shared components (enums and interfaces).

What about mixins -- are those a shared component, or does this component expect the app to provide the necessary mixins?

src/interfaces/address-interface.ts
```diff
@@ -0,0 +1,31 @@
+/** Interface to define a base address. */
```
severinbeauvais, line 1: Maybe pluralize this file name so we can tell there are multiple IFs in here.

src/components/ShareStructure/EditShareStructure.vue
severinbeauvais: If this component expects the app to provide the CurrencyLookupMixin, I think we should start adding comments to that effect (eg, line 176). Maybe...

`// this component expects this mixin/interface/enum/blah to exist in the app`

or something to the effect that it's not importing its own dependencies from the shared library.

src/mixins/index.ts
```diff
@@ -1,7 +1,11 @@
+import CommonMixin from './common-mixin'
+import CurrencyLookupMixin from './currency-lookup-mixin'
```
severinbeauvais, line 2: So currency-lookup-mixin is not exported/published ?

src/modules/corp-type-module/corp-type-module.ts
```diff
@@ -101,7 +101,7 @@ const CorpInfoArray: Array<CorpInfoIF> = [
     colinInd: true,
     corpClass: CorpClass.BC,
     shortDesc: 'BENEFIT COMPANY',
-    fullDesc: 'Benefit Company',
+    fullDesc: 'BC Benefit Company',
```
severinbeauvais, line 104: Thanks.

I guess multiple dependencies will need to be updated once this change is published. Please keep a todo list (or create tickets for that).

severinbeauvais (approved):
I didn't test this locally but it looks OK (except for comments).

GTG when you're happy with it.

package.json
```diff
@@ -56,7 +58,7 @@
     "@vue/cli-service": "4.2.3",
     "@vue/eslint-config-standard": "^4.0.0",
     "@vue/eslint-config-typescript": "^4.0.0",
-    "@vue/test-utils": "1.0.0-beta.28",
+    "@vue/test-utils": "1.0.0-beta.33",
```
severinbeauvais, line 61: Did you try the latest version?
https://www.npmjs.com/package/@vue/test-utils

It would be nice to graduate past the alpha/beta versions :)

(Yes, version 2 is in beta. I'm OK with staying on v1 for now.)

src/components/ShareStructure/ShareStructure.vue
```diff
@@ -0,0 +1,1058 @@
...
+import { v4 as uuidv4 } from 'uuid'
+
+// Components
+import { ActionChip } from '@/components/ActionChip'
+import { ConfirmDialog } from '@/components/ConfirmDialog'
+import EditShareStructure from './EditShareStructure.vue'
```
severinbeauvais, line 405: As long as ShareStructure.vue and EditShareStructure.vue are packaged/published together, this is fine/good.

src/components/ShareStructure/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as ShareStructure } from './ShareStructure.vue'
+export { default as EditShareStructure } from './EditShareStructure.vue'
```
severinbeauvais, line 2: Why does Edit have to be here? Does the app (or Storybook) need to import it?

src/components/ConfirmDialog/pacakge.json
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
cameron-freshworks: fixed*

src/components/ConfirmDialog/ConfirmDialog.stories.ts
cameron-freshworks: A cool little trick provided by the internet!

src/components/ShareStructure/ShareStructure.vue
cameron-freshworks: Good catch, will do

src/interfaces/address-interface.ts
cameron-freshworks: k, will do

src/components/ShareStructure/ShareStructure.vue
severinbeauvais: I'm ** hoping ** that won't cause another (nested) copy of lodash in the node_modules folder...

src/mixins/index.ts
cameron-freshworks: I don't think we've published any mixins yet? 
Is that something we should do? 
Do you feel CurrencyLookup would be a good one to use elsewhere?

src/modules/corp-type-module/corp-type-module.ts
cameron-freshworks: Definitely 👍

src/components/ShareStructure/ShareStructure.vue
severinbeauvais: If it does... Maybe we abandon the "self sufficient shared component" idea and just rely on the app's dependencies -- but add a comment to that effect in the code here.

Alternately, if we must have self-sufficient components then you could import just the lodash methods you need (eg, https://www.npmjs.com/package/lodash.clonedeep etc) instead of the whole library.

package.json
cameron-freshworks: I didn't try no, was content with a version i knew works well with the ported unit testing.
I will try 1.1.3 though, to see if we can officially graduate.

src/components/ShareStructure/index.ts
cameron-freshworks: Added it for the unit testing

cameron-freshworks: Storybook is ok without this

src/components/ShareStructure/ShareStructure.vue
severinbeauvais: > just rely on the app's dependencies

A slippery slope. How much should this component add to its package.json and how much should it depend on the app? All? None? I don't have an answer but we should try to answer this.

src/mixins/index.ts
severinbeauvais: I think you're right. We only have some partial mixins in here.

I think, ultimately, it would make sense to publish some mixins (ie, useful code that multiple apps could use). But it will depend on what they depend on (prob no store, depends on imports, etc).

As for CurrencyLookup, I guess you're only using it here, right? I'd be tempted to move it into the component's folder to keep it close to where it's used (and the whole thing gets published together).

Or it could the first mixin (or is that module -- see corp-type-module) that is published.

Currently, does this mixin have to exist in the app?

src/components/ShareStructure/index.ts
severinbeauvais: Weird that unit tests don't import ShareStructure which imports EditShareStructure.

src/components/ActionChip/ActionChip.vue
cameron-freshworks: I'm not entirely sure on this, i believe we will only have issues if there is a conflict with an already existing file in the same project. 
I do agree though, will update these imports.

src/components/ShareStructure/EditShareStructure.vue
cameron-freshworks: Good point, i think if we were relying on something the app would be providing.
Is the comment needed when the import is contained within the shared repo?
Doesn't the import already imply that?

src/mixins/index.ts
cameron-freshworks: I feel like we should publish it, if we find we need it.
Yes, i believe we need this mixin, unless you would prefer to see that function within the component itself. 
I prefer the mixin 🤷

package.json
cameron-freshworks: Updated 👍

src/components/ActionChip/ActionChip.vue
cameron-freshworks: These import changes actually break the build. 
I think these changes will create a little tech debt, 
We could leave this imports BUT this will require to clear the conflicting enum/interface out of repo where these are pulled in potentially. 
I'm not sure the best answer. I'd like to see the conflict here though so perhaps we can better solve this.

src/components/ShareStructure/EditShareStructure.vue
severinbeauvais: > Good point, i think if we were relying on something the app would be providing.
👍 

> Is the comment needed when the import is contained within the shared repo?
> Doesn't the import already imply that?

Right. I think it's obvious then. Only when the shared component expects the app to provide the local thing.

src/mixins/index.ts
severinbeauvais: I was thinking this mixin could be moved into the component's folder (which, I know, isn't something we've done before, but this feels right here).

Up to you... wfm!

src/components/ShareStructure/ShareStructure.vue
severinbeauvais: Per chat in RC, we can determine this per scenario.

severinbeauvais: > I'm ** hoping ** that won't cause another (nested) copy of lodash in the node_modules folder...

Does it? 😬

cameron-freshworks: I'm not sure how to tell. 
By including it here, it will include it in the Node_Modules for the Shared Components. 
Is that the one you are referring too?

src/interfaces/address-interface.ts
cameron-freshworks: Updated 👍

src/mixins/index.ts
cameron-freshworks: okie doke, will relocate 👍 
Updated in next commit.

src/components/ShareStructure/ShareStructure.vue
severinbeauvais: I'm hoping there aren't multiple, full copies of lodash...
node_modules/lodash
node_modules/bcrs-shared-components/your_component/lodash
and others...

cameron-freshworks: ah i see. 
I suspect we will have..

node_modules/lodash
node_modules/bcrs-shared-components/your_component/lodash

Thoughts on how to resolve this?

cameron-freshworks: I'm not sure how to do this.. 

Is there a way to make it a dependency of the root app, if they pull in this component IF they don't already have it?

cameron-freshworks: Or we just pull in the specific functions in the shared comp.
 I believe you suggested that idea already.

package.json
```diff
@@ -56,7 +58,7 @@
     "@vue/cli-service": "4.2.3",
     "@vue/eslint-config-standard": "^4.0.0",
     "@vue/eslint-config-typescript": "^4.0.0",
-    "@vue/test-utils": "1.0.0-beta.28",
+    "@vue/test-utils": "^1.1.3",
```
severinbeauvais, line 61: 🎉

src/components/ShareStructure/index.ts
severinbeauvais, line 2: Do you need to also include the currency lookup mixin here?

severinbeauvais: Or not, if it's not needed anywhere else atm?

severinbeauvais (approved):

src/components/ShareStructure/ShareStructure.vue
severinbeauvais: > Is there a way to make it a dependency of the root app, if they pull in this component IF they don't already have Lodash?

I don't think so. But the developer pulling it in will surely see the broken build.

Argh, I don't know which I prefer (or hate the least)... If we depend on the parent then what if they provide an incompatible version of the import? (bad parenting, haha) Dunno. Keep it in mind, decide later.

src/components/ShareStructure/index.ts
cameron-freshworks: It isn't used anywhere else atm, so i believe we are good without exporting for the time being.

PR #29, opened 2021-03-08
Certify component wording and staff section

jordiwes:
Issue #: /bcgov/entity#5745

Description of changes:
Changes to certify component as per:
https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/5475
Some wording changes and ability to have staff fill out


src/components/Certify/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@bcrs-shared-components/certify",
-    "version": "1.0.6",
+    "version": "1.0.7",
```
severinbeauvais, line 3: No, you don't do this. Lerna does that in step 14.

https://github.com/bcgov/bcrs-shared-components/blob/master/README.md

You may not have sufficient permissions to do steps 14 and 15. Just ask Cam or me to do it for you.

tests/unit/Certify.spec.ts
```diff
@@ -118,6 +120,13 @@ describe('Certify', () => {
     expect(getLastEvent(wrapper, 'valid')).toBe(true)
   })
 
+  it('is valid when both certifiedBy, isCertified, and is staff are defined', () => {
+    const wrapper: Wrapper<Certify> = createComponent(trimmedCertifier, true, true)
+
+    // The last "valid" event should indicate that the form is valid.
+    expect(getLastEvent(wrapper, 'valid')).toBe(true)
+  })
+
```
severinbeauvais, line 129: How about a test that it's invalid when staff is True but the other 2 values aren't True?

jordiwes: Done

src/components/Certify/package.json
jordiwes: Ah oh gotcha. changed it back

severinbeauvais (approved):
Code looks good.

Please commit the updated Storybook build files.

PR #30, opened 2021-03-12
Import fixes and SummaryMode for ShareStructure

cameron-freshworks:
*Issue #:* /bcgov/entity#5178
*Issue #:* /bcgov/entity#6172

*NOTE:* - This work is a dependency of me completing https://github.com/bcgov/business-edit-ui/pull/125

*Description of changes:*

* Fixes some import names due to directory name change after publishing of packages.

* Relocate/rename a mixin only used by ShareStructure.

* Implements a non editable mode of ShareStructure.

* Updates unit testing.


src/components/ShareStructure/ShareStructure.vue
```diff
@@ -216,12 +217,13 @@
         <!-- Share Series rows -->
         <template v-for="(seriesItem, index) in row.item.series">
           <tr
-            v-if="showSeriesEditForm[row.index] && !showSeriesEditForm[row.index][index]"
+            v-if="showSeriesEditForm[row.index] && !showSeriesEditForm[row.index][index] && !(!isEditMode &&
+            (seriesItem.action === ActionTypes.REMOVED || row.item.action === ActionTypes.REMOVED))"
```
severinbeauvais, line 221: I recommend a helper function here.

src/components/ShareStructure/share-mixin.ts
```diff
@@ -1,25 +1,15 @@
 import { Component, Vue } from 'vue-property-decorator'
 
 /**
- * Mixin that provides some useful common utilities.
+ * Mixin that provides some utilities.
  */
 @Component({})
-export default class CommonMixin extends Vue {
+export default class ShareMixin extends Vue {
```
severinbeauvais, line 7: I like it.

severinbeauvais: Can you update the comment (line 4) to say what this file is -- why it looks like Common Mixin but isn't?

tests/unit/ConfirmDialog.spec.ts
```diff
@@ -2,7 +2,7 @@ import Vue from 'vue'
 import Vuetify from 'vuetify'
 import Affix from 'vue-affix'
 import { mount } from '@vue/test-utils'
-import { ConfirmDialog } from '@/components/ConfirmDialog'
+import { ConfirmDialog } from '@/components/confirm-dialog'
```
severinbeauvais, line 5: I'm confused as to how this works. As far as this unit test is concerned, ConfirmDialog is a local component in `@/components/ConfirmDialog`, isn't it?

(Same comment for ActionChip in ShareStructure.spec.ts.)

severinbeauvais (approved):
Just some small comments. GTG when you're happy with it.

src/components/ShareStructure/ShareStructure.vue
cameron-freshworks: Will do 👍

tests/unit/ConfirmDialog.spec.ts
cameron-freshworks: The issue i'm having with our directory structure, is after we publish these packages and import them. The directory name is converted from `ActionChip` to `action-chip`.

image

So when you have a component that utilizes another shared component, the import path is wrong in the component on the side of implementation. 

So i just renamed the directory to match the name it's converted too anyways. ie `action-chip` etc. I'm not sure of a way around this, but this conversion is an issue.

cameron-freshworks: So we can maybe go two ways here...
Update all the directory names to the format they become on the side of integration 
OR
figure out why/how these conversions happen and handle it.

Either way, it is a bit of tech debt.

PR #31, opened 2021-03-16
Ux share structure

cameron-freshworks:
*Issue #:* /bcgov/entity#5178

*Description of changes:*

- UX updates to Share Structure


src/interfaces/package-lock.json
```diff
@@ -0,0 +1,66 @@
+{
```
severinbeauvais, line 1: That's weird -- how did this file get created? I don't see package-lock files for any other component in this repo.

Either all shared components should have a package-lock file, or none should (I think).

cameron-freshworks: ahh, perhaps during the hoist. Will clean these up

severinbeauvais (approved):
GTG when you're happy with it.

src/interfaces/package-lock.json
cameron-freshworks: Done 👍

PR #32, opened 2021-03-17
Updated ShareStructure component

severinbeauvais:
*Issue #:* /bcgov/entity#5175

*Description of changes:*
- updated Share Structure component to use share structure data instead of business snapshot
- fixed unit test warnings
- rebuilt Storybook


src/components/ShareStructure/ShareStructure.vue
```diff
@@ -434,30 +434,30 @@ export default class ShareStructure extends Mixins(ShareMixin) {
   // Props
   /** Edit Mode */
   @Prop({ default: true })
-  private isEditMode!: boolean
+  readonly isEditMode!: boolean
 
-  @Prop()
-  private isCorrection: boolean
+  @Prop({ default: false })
```
severinbeauvais, line 439: Default before and now are both falsy.

src/components/ShareStructure/ShareStructure.vue
```diff
@@ -434,30 +434,30 @@ export default class ShareStructure extends Mixins(ShareMixin) {
...
+  readonly isCorrection: boolean
 
   @Prop({ default: null })
-  private incorporationApplication: IncorporationFilingIF
+  readonly incorporationApplication: IncorporationFilingIF
 
   @Prop({ default: null })
-  private businessSnapshot: BusinessSnapshotIF[]
+  readonly originalShareStructure: ShareStructureIF
```
severinbeauvais, line 446: Passing in just the Share Structure, so that THIS component doesn't need to know about the business snapshot.

src/components/ShareStructure/ShareStructure.vue
```diff
@@ -471,7 +471,6 @@ export default class ShareStructure extends Mixins(ShareMixin) {
   private showSeriesEditForm: Array<any> = this.mapEmpty2dArray()
   private addEditInProgress = false
   private currentShareStructure: ShareClassIF = null
-  private originalSnapshot: BusinessSnapshotIF[]
```
severinbeauvais, line 474: I also deleted this IF from the shared library as it's no longer being used.

src/components/ShareStructure/ShareStructure.vue
```diff
@@ -614,9 +612,9 @@ export default class ShareStructure extends Mixins(ShareMixin) {
    * @params shareClass The Share class to compare
    */
   private isShareClassEdited (shareClass: ShareClassIF): boolean {
-    const originalShareClasses: any = this.isCorrection
+    const originalShareClasses = this.isCorrection
       ? cloneDeep(this.originalIA.incorporationApplication.shareStructure.shareClasses)
-      : cloneDeep(this.originalSnapshot[4].shareClasses)
```
severinbeauvais, line 619: I could see no reason to case the type to `any`. In this case the type is ShareClassIF.

Same below.

src/components/ShareStructure/ShareStructure.vue
```diff
@@ -921,11 +919,6 @@ export default class ShareStructure extends Mixins(ShareMixin) {
     return new Array(50).fill(null).map(() => new Array(50).fill(null))
   }
 
-  @Watch('businessSnapshot')
-  private updateSnapshot (): void {
-    this.originalSnapshot = cloneDeep(this.businessSnapshot)
```
severinbeauvais, line 926: No need to watch the prop and copy to a local variable. Instead, the script just uses the prop directly.

src/components/ShareStructure/ShareStructure.vue
```diff
@@ -522,7 +521,6 @@ export default class ShareStructure extends Mixins(ShareMixin) {
   // Initialize data sets
   mounted () {
     this.updateOriginalIA()
```
severinbeauvais, line 523: FYI, instead of updating the local variable here, adding "immediate" to the watch might do the same thing (ie, fire the watch method when this component is initialized).

tests/unit/ShareStructure.spec.ts
```diff
@@ -328,13 +328,13 @@ describe('Share Structure component', () => {
   })
 
   it('checks for the Action chips on modified Class shares', async () => {
-    expect(wrapper.find(ActionChip).exists()).toBe(true)
+    expect(wrapper.findComponent(ActionChip).exists()).toBe(true)
```
severinbeauvais, line 331: All these changes are to handle deprecated unit test method warnings

src/components/ShareStructure/ShareStructure.vue
cameron-freshworks: Ok 👍

cameron-freshworks (approved):
Looks good!

PR #35, opened 2021-03-18
Ux Updates to CourtOrder & Share Structure

cameron-freshworks:
*Issue #:* /bcgov/entity#6413

*Description of changes:*

* Updates POA for validations.

* Minor UX / Style adjustments for Sharestructure.

* Storybook Build


src/components/CourtOrderPoa/CourtOrderPoa.vue
```diff
@@ -50,9 +52,20 @@ export default class CourtOrderPoa extends Vue {
   private planOfArrangement = false
   private valid = false
 
-  @Watch('validate')
+  /** Clear rules and reset validations. */
+  private clearValidations (): void {
+    this.courtOrderNumRules = []
+    this.$refs.courtNumRef.resetValidation()
```
severinbeauvais, line 58: You didn't need to update the local IF declaration for this to work?

BTW, I like this validation design.

src/components/CourtOrderPoa/CourtOrderPoa.vue
```diff
@@ -50,9 +52,20 @@ export default class CourtOrderPoa extends Vue {
...
+  }
+
+  /** Local getter to know when to validate. */
+  private get validateForm (): boolean {
+    return this.validate && this.planOfArrangement
```
severinbeauvais, line 63: Check the form API method... they may be async, or need an `await nextTick()` after them before correct results are available.

I just remember that the DOM isn't updated right away (although the result from `validate` is correct). I'm not sure if you have to do anything here (or where you call resetValidation())...

I needed to do this to check the state (may not apply to you):
https://github.com/bcgov/business-edit-ui/blob/13ced0a00ddc493645b16967a2a4fe1c5aca2649/src/components/common/EffectiveDateTime.vue#L388

severinbeauvais (approved):
GTG when you're happy with this.

src/components/CourtOrderPoa/CourtOrderPoa.vue
cameron-freshworks: Hmm, i'll look.
Thanks!

src/components/CourtOrderPoa/CourtOrderPoa.vue
cameron-freshworks: kk, i'll look into this 👍

src/components/CourtOrderPoa/CourtOrderPoa.vue
cameron-freshworks: Ah right, it already existed :)

src/components/CourtOrderPoa/CourtOrderPoa.vue
cameron-freshworks: I tried both, but it seems to the same. 
I believe once file and pay is hit, the `validate` is constant and will then only trigger based on the local value of `planOfArrangement` (which is also kept in sync app level). 

I'm sure this could be improved (as always) but lets go with this for now :)

cameron-freshworks: TLDR: not noticing anything funky without awaiting 😆

severinbeauvais: Also, I'd be tempted to make this a function instead of a getter.
1. so it's consistent with clearValidation() and others
2. in case it gets called lot (as getters sometimes do)
YMMV

PR #36, opened 2021-03-18
POA Column fix

cameron-freshworks:
*Issue #:* /bcgov/entity#6413

* Fixes columns to bring in line with other summary components columns.
* Rebuild.


src/components/CourtOrderPoa/CourtOrderPoa.vue
```diff
@@ -1,10 +1,10 @@
 <template>
   <v-card flat class="mt-4 pa-8">
     <v-row no-gutters align="start">
-      <v-col cols="3">
+      <v-col cols="2">
```
severinbeauvais, line 4: The other Edit UI components use 3/9 for columns. Could it the v-card margin/padding throwing things off?

cameron-freshworks: The Documents and Certify are at 2, i suppose we should fix those then.

PR #37, opened 2021-03-19
5475 more certify UI updates

jordiwes:
Issue #: /bcgov/entity#5475

Description of changes:

- UI updates to match comp (mostly font weights and colors).
See https://projects.invisionapp.com/share/42105KY639E7#/screens/444385469
- Changed v-flex to v-row and v-col
- Added first and second column widths as props
- Storybook build.


severinbeauvais (approved):
LGTM.
Need to fix a conflict though.

cameron-freshworks (commented):
Looks good, one question though. The column size here on a regular laptop is shrinking to a 2 col but i think we want 3 everywhere, maybe @severinbeauvais would know about this as well?


Screen Shot 2021-03-19 at 12 04 05 PM

severinbeauvais:
> Looks good, one question though. The column size here on a regular laptop is shrinking to a 2 col but i think we want 3 everywhere, maybe @severinbeauvais would know about this as well?
> 
> Screen Shot 2021-03-19 at 12 04 05 PM

1. We need to stop using v-flex as it's deprecated. It's easy to convert to v-row and v-col.
2. We haven't had a good look at responsiveness (different screen widths), so kudos for doing it here. At some point the other components should be looked at. Wanna do that now?

cameron-freshworks:
> > Looks good, one question though. The column size here on a regular laptop is shrinking to a 2 col but i think we want 3 everywhere, maybe @severinbeauvais would know about this as well?
> > Screen Shot 2021-03-19 at 12 04 05 PM
> 
> 1. We need to stop using v-flex as it's deprecated. It's easy to convert to v-row and v-col.
> 2. We haven't had a good look at responsiveness (different screen widths), so kudos for doing it here. At some point the other components should be looked at. Wanna do that now?

I know from experience going back and forth between Shared and the UI where the components are used can be a pain in the butt :) I would try and do it now, to save the hassle later!

jordiwes:
So I changed to 3 & 9 columns vs 2 & 10 to match what Severin had in the alteration date & time, but not sure if there is too much white space now? Might be fine:

image

src/components/Certify/Certify.vue
```diff
@@ -1,21 +1,28 @@
...
-            <v-text-field
-              filled
-              persistent-hint
-              id="certified-by-textfield"
-              label="Legal name of authorized person"
-              :value="certifiedBy"
-              :rules="[(v) => !!v || 'A person\'s legal name is required.']"
-              @input="emitCertifiedBy($event)"
-            />
+      <v-container>
```
severinbeauvais, line 4: Do you need the container? (I'm not sure.)

src/components/Certify/Certify.vue
```diff
@@ -1,21 +1,28 @@
...
+        <v-row class="pl-4">
+          <v-col cols="3" md="3" class="firstCol">
+            <label><strong>Legal Name</strong></label>
+          </v-col>
+          <v-col cols="9" md="9" class="secondCol">
```
severinbeauvais, line 9: I'm not sure if you need `md="9"` since it's the same as the cols value. Please remove if not needed.

src/components/Certify/Certify.vue
```diff
@@ -159,6 +166,10 @@ export default class Certify extends Vue {
   color: $gray7;
   font-weight: normal;
 }
+.firstCol, .secondCol {
+    padding-left: 0px;
+    padding-right: 0px;
```
severinbeauvais, line 171: I have found that it's often easier to have these layout tweaks right in the template (HTML) -- especially when one has to update a bunch of layout on a page. In this case you can use the built-in Vuetify classes to set the padding: `px-0`.

severinbeauvais:
> So I changed to 3 & 9 columns vs 2 & 10 to match what Severin had in the alteration date & time, but not sure if there is too much white space now? Might be fine:

What does it look like in use (in app) ? Hopefully our UI designs will be consistent w.r.t. label and content spacing, ie. 3+9 for everything. Either way, have a look it and make sure it matches the comps.

Another thing you could do (but only if needed) is to pass in the first/second columns width as props (with default 3 & 9). This would allow this component to work in various apps (if they have different requirements).

src/components/Certify/Certify.vue
jordiwes: I made this change

src/components/Certify/Certify.vue
jordiwes: removed

jordiwes:
I made the columns props, as I know corrections use 2 & 10

severinbeauvais:
@jordiwes Please use the standard template for this PR's description block... and then I will merge this change.

PS Let me know if you need help with the rest of the shared components steps.

jordiwes:
Done

severinbeauvais:
Oh no. I just noticed you added some commits after my review, and they are incorrect (and I merged them).

It appears you did not follow the steps in the [README](https://github.com/bcgov/bcrs-shared-components/blob/master/README.md).

You do not update the component versions. Lerna does that for you.

Try to reconcile what you did with the steps in the readme and continue from there. You may or may not have sufficient permissions to push tage or publish -- if not, let me know and I will do that for you.

PR #53, opened 2021-04-21
Changes from UX assurance that couldn't be done with just styling

jachurchill:
*Issue #:* /bcgov/entity#6378

*Description of changes:* Changes from UX assurance that couldn't be done with just styling

### DIFF
--- a/src/components/StaffPayment/StaffPayment.vue
+++ b/src/components/StaffPayment/StaffPayment.vue
@@ -13,6 +13,7 @@
               filled
               id="routing-slip-number-textfield"
               label="Routing Slip Number"
+              hint="FAS Routing Slip Number (9 digits)"
               :value="staffPaymentData.routingSlipNumber"
               :rules="routingSlipNumberRules"
               :disabled="paymentOption === StaffPaymentOptions.BCOL || paymentOption === StaffPaymentOptions.NO_FEE"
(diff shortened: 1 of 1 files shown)


src/components/StaffPayment/StaffPayment.vue
```diff
@@ -59,6 +60,8 @@
           <!-- No Fee radio button -->
           <v-radio id="no-fee-radio" label="No Fee" :value="StaffPaymentOptions.NO_FEE" />
 
+          <hr>
```
cameron-freshworks, line 63: What's the difference between a <br> and <hr>?

cameron-freshworks (commented):
Just one comment about the divider, curious if the styling is different than other v-dividers we've used elsewhere in the project.

src/components/StaffPayment/StaffPayment.vue
severinbeauvais: It looks to me like they're different.
https://www.w3schools.com/howto/howto_css_style_hr.asp
https://vuetifyjs.com/en/components/dividers/

Therefore I think we should use `v-divider` here for consistency.

severinbeauvais (commented):
In additional to my comment above, Cam or I will need to pick this up as John has left now.

cameron-freshworks:
> In additional to my comment above, Cam or I will need to pick this up as John has left now.

kk, no prob. I'll pull these branches and support them through!

severinbeauvais:
> > In additional to my comment above, Cam or I will need to pick this up as John has left now.
> 
> kk, no prob. I'll pull these branches and support them through!

Thanks. If you get busy with something else, lmk and I can take it on.

cameron-freshworks:
Closing this PR due to the work done in PR https://github.com/bcgov/bcrs-shared-components/pull/54

PR #55, opened 2021-04-22
5693 Prevent ShareClass Removal

cameron-freshworks:
*Issue #:* /bcgov/entity#5693

*Description of changes:*

* Added error text and prop for invalid minimum share classes.
* Updated Tests.

*Note:*
* The logic that dictates the actual state of validity is within the Edit UI 👍 


src/components/ShareStructure/ShareStructure.vue
```diff
@@ -476,6 +488,10 @@ export default class ShareStructure extends Mixins(ShareMixin) {
   @Prop({ default: false })
   readonly invalidSection!: boolean
 
+  /** Verification the Share Structure contains the minimum required Share Classes. */
+  @Prop({ default: false })
+  readonly invalidMinimumShareClass!: boolean
```
severinbeauvais, line 493: I think this should be generalized so it can be reused for possible future additional checks.

Also, this component doesn't really "know" what the invalid condition is, nor does it (or should it) care, which is another reason to rename this.

cameron-freshworks: 🤔 I thought about that too but the styling effect it has on the component is specific. 
ie the invalidSection and invalidMinimum.. have different style outcomes in the UI. 
So as long as our designs call for specific messages or styling, i think we need to be specific no?

severinbeauvais: You're right about the message, but the styling is generic right? I guess you could pass in the message too.

cameron-freshworks: Not exactly generic. 
The css is for sure but where it applies is not, ie different scenarios call for different places to apply the styling. 
We could pass in the message, but it's just another prop to handle and until we have a requirement for multiple messages i think it's ok to have a single message maybe?

severinbeauvais: OK, leave it as is. Yes, we can review this if we need another message. No problem!

severinbeauvais (approved):

src/components/ShareStructure/ShareStructure.vue
cameron-freshworks: Sounds good, thanks. Good discussion though, as always :)

PR #60, opened 2021-05-05
6724 - DatePicker disabled style

andrepestana-aot:
*Issue #:* /bcgov/entity#6724

*Description of changes:*

- Fixed DatePicker disabled style


src/components/DatePicker/DatePicker.vue
```diff
@@ -163,4 +163,16 @@ export default class DatePicker extends Vue {
 ::v-deep .v-icon.v-icon.v-icon--link {
   cursor: text;
 }
+
+::v-deep .theme--light.v-icon.v-icon.v-icon--disabled {
```
severinbeauvais, line 167: Do these need to be v-deep, since you're not reaching in to a rendered component now?

severinbeauvais: PS I doubt that `.theme--light` is useful here and in selectors below.

andrepestana-aot: Yes, we do need v-deep because it is inside v-text-field.

image
Maybe for the other I can get rid of .theme--light but for line 167 if it doesn't match exactly the selector, it doesn't override it.

andrepestana-aot: I fixed one css selector but the other 2 need the `.theme--light`

severinbeauvais (approved):

cameron-freshworks (approved):

cameron-freshworks:
Keep in mind this will affect all instances of this component and if that was intended by design, all good :) 

ie Resolution Dates in Articles and Resolution Dates Dialog 👍

severinbeauvais:
@argush3 Do you want to have a look at this or can we go ahead and merge it?

andrepestana-aot:
> Keep in mind this will affect all instances of this component and if that was intended by design, all good :)
> 
> ie Resolution Dates in Articles and Resolution Dates Dialog 👍

Yes, I checked and they don't have the disabled state in any scenario. Btw this is how the component should behaviour in all cases when disabled. Thanks!

argush3 (approved):
LGTM 👍

PR #61, opened 2021-05-06
Added Folio Number component

severinbeauvais:
*Issue #:* /bcgov/entity#5024

*Description of changes:*
- added Folio Number component
- updated Storybook


src/components/FolioNumber/FolioNumber.vue
```diff
@@ -0,0 +1,264 @@
...
+
+  /** Informs parent whether we are in editing mode. */
+  @Watch('isEditing')
+  @Emit('isEditing')
+  private emitIsEditing (val: boolean): void {}
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
```
cameron-freshworks, line 228: No action required here!

Just sharing a thought, in the Name Request UI when we import these components it throws errors because this is pulling the `assets` from root. It works good in create/edit where the assets file structure is the same but in Name Request, one of the sub directories is different. 

That being said, down the road we will need to figure this one out, maybe using an external style sheet that needs to be a dependency?? i'm not sure 🤷

cameron-freshworks (approved):
Looks good to me!

PR #64, opened 2021-05-14
Updated interfaces to import other interfaces locally

severinbeauvais:
*Issue #:* /bcgov/entity#6554

*Description of changes:*
- updated interfaces to import other interfaces locally
- there are no Storybook changes for this code change

### DIFF
--- a/src/interfaces/incorporation-application-interface.ts
+++ b/src/interfaces/incorporation-application-interface.ts
@@ -1,5 +1,4 @@
-import { IncorporationAddressIf, ShareClassIF,
-  NameTranslationIF } from '@/interfaces'
+import { IncorporationAddressIf, ShareClassIF, NameTranslationIF } from './'
 
 /** Incorporation Application filing loaded from / saved to the Legal API. */
 export interface IncorporationApplicationIF {
--- a/src/interfaces/incorporation-filing-interface.ts
+++ b/src/interfaces/incorporation-filing-interface.ts
@@ -1,4 +1,4 @@
-import { IncorporationApplicationIF } from '@/interfaces'
+import { IncorporationApplicationIF } from './'
 
 /** Incorporation Application filing loaded from / saved to the Legal API. */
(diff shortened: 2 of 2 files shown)


src/interfaces/incorporation-application-interface.ts
```diff
@@ -1,5 +1,4 @@
-import { IncorporationAddressIf, ShareClassIF,
-  NameTranslationIF } from '@/interfaces'
+import { IncorporationAddressIf, ShareClassIF, NameTranslationIF } from './'
```
severinbeauvais, line 1: The previous code had this interface looking in the parent's interfaces folder for these imports. This is incorrect. This interface should look in THIS folder for any imports.

Any shared interface (or mixin, enum, etc) should be able to get everything it needs from the current folder structure (this repo).

It's slightly different if an interface needs an enum... In this example, the interface should import from "@bcrs-shared-components/enums" so that it gets the versioned one.

cameron-freshworks (approved):
Looks good 👍

severinbeauvais:
So, a bunch of components were updated as a result of this interface import change:
 - @bcrs-shared-components/certify@1.0.19
 - @bcrs-shared-components/contact-info@1.0.14
 - @bcrs-shared-components/court-order-poa@1.0.9
 - @bcrs-shared-components/fee-summary@1.1.2
 - @bcrs-shared-components/share-structure@1.0.15
 - @bcrs-shared-components/staff-payment@1.0.23
 - @bcrs-shared-components/action-chip@1.0.7
 - @bcrs-shared-components/interfaces@1.0.20

In my current ticket (6554), I only need the updated interfaces (1.0.20) and staff-payment (1.0.23).

I think no other dependenies (in any app) need to be updated at this time, since they obviously didn't experience the import issue I did. If and when they need an updated component, then at that time they can update their dependency(ies) (and do the appropriate testing, etc).

PR #66, opened 2021-05-14
4388 - Updated CourtOrderPoa layout and fixed validation 

andrepestana-aot:
*Issue #:* /bcgov/entity#4388

*Description of changes:*

- Added displaySideLabels props;
- Fixed validation when POA is not checked;
- Updated tests.


src/components/CourtOrderPoa/CourtOrderPoa.vue
```diff
@@ -44,7 +44,7 @@ export default class CourtOrderPoa extends Vue {
 
   /** Prompt the validations. Used for global validations. */
   @Prop({ default: false })
-  private validate
```
andrepestana-aot, line 47: I needed to change the prop's name because it conflicted with FormIF

severinbeauvais: It wouldn't hurt to add a type to this prop.

Also, props should be readonly (not private).

src/components/CourtOrderPoa/CourtOrderPoa.vue
```diff
@@ -76,22 +80,26 @@ export default class CourtOrderPoa extends Vue {
...
-        (v: string) => !/\s$/g.test(v) || 'Invalid spaces', // trailing spaces
-        (v: string) => !(v.length < 5) || 'Court order number is invalid',
-        (v: string) => !(v.length > 20) || 'Court order number is invalid'
+        (v: string) => (!v || !/^\s/g.test(v)) || 'Invalid spaces', // leading spaces
+        (v: string) => (!v || !/\s$/g.test(v)) || 'Invalid spaces', // trailing spaces
+        (v: string) => (!v || !(v.length < 5)) || 'Court order number is invalid',
+        (v: string) => (!v || !(v.length > 20)) || 'Court order number is invalid'
       ]
+      if (this.planOfArrangement) {
+        this.courtOrderNumRules.splice(0, 0, (v: string) => (!!v && this.planOfArrangement) || 'A Court Order number is required')
```
severinbeauvais, line 101: Since courtOrderNumRules is an array, you might find it cleaner to just push another function into it :)

tests/unit/CourtOrderPoa.spec.ts
```diff
@@ -98,26 +112,29 @@ describe('Court Order and Plan of Arrangement component', () => {
...
-    await Vue.nextTick()
-
-    // Prompt validates through prop
-    wrapper.setProps({ validate: true })
-    await Vue.nextTick()
-
+    // Verify it continues invalid
     expect(wrapper.find('#court-num-form').text()).toContain('Court order number is invalid')
-    expect(wrapper.emitted('emitValid').pop()[0]).toEqual(false)
+    expect(wrapper.emitted().emitValid.length).toBe(1)
```
severinbeauvais, line 117: Checking the number of events emitted is good, but you should definitely verify that the last event has the correct value.

tests/unit/CourtOrderPoa.spec.ts
```diff
@@ -57,9 +63,6 @@ describe('Court Order and Plan of Arrangement component', () => {
 
     // Verify checkbox is selected
     expect(wrapper.vm.$data.planOfArrangement).toBe(true)
-
-    // Prompt validates through prop
-    wrapper.setProps({ validate: true })
     await Vue.nextTick()
```
severinbeauvais, line 66: Do you still need the await here?

tests/unit/CourtOrderPoa.spec.ts
```diff
@@ -127,17 +144,9 @@ describe('Court Order and Plan of Arrangement component', () => {
...
-    await Vue.nextTick()
-
-    // Prompt validates through prop
-    wrapper.setProps({ validate: true })
-    await Vue.nextTick()
-
+    // Verify it continues invalid
     expect(wrapper.find('#court-num-form').text()).toContain('Court order number is invalid')
-    expect(wrapper.emitted('emitValid').pop()[0]).toEqual(false)
+    expect(wrapper.emitted().emitValid.length).toBe(1)
```
severinbeauvais, line 149: Verify value here, too.

src/components/CourtOrderPoa/CourtOrderPoa.vue
andrepestana-aot: Yes, you're right. It should be public. 👍

src/components/CourtOrderPoa/CourtOrderPoa.vue
andrepestana-aot: Yes. At the beginning I was thinking of the order of the rules but now it doesn't matter anymore. 👍

tests/unit/CourtOrderPoa.spec.ts
andrepestana-aot: It is checked on line 105.

tests/unit/CourtOrderPoa.spec.ts
andrepestana-aot: Checked on line 137.

tests/unit/CourtOrderPoa.spec.ts
andrepestana-aot: Yes, you're right! Thanks!

src/components/CourtOrderPoa/CourtOrderPoa.vue
severinbeauvais: Well, it doesn't need to be public since nothing external needs to use this class variable.

src/components/CourtOrderPoa/CourtOrderPoa.vue
severinbeauvais: Good point.

tests/unit/CourtOrderPoa.spec.ts
severinbeauvais: Ok, I see now (and your code comment - thanks) You can always check it again. Up to you.

src/components/CourtOrderPoa/CourtOrderPoa.vue
andrepestana-aot: Expose or not expose? 🤔

severinbeauvais: My point is it should be readonly.

As it turns out, public/private doesn't work in TS at the moment anyway so 🤷‍♂️

severinbeauvais (approved):
Fix anything if needed, but you have approval.

src/components/CourtOrderPoa/CourtOrderPoa.vue
andrepestana-aot: ok. Sounds good.

PR #70, opened 2021-05-27
Misc tweaks to Staff Payment

severinbeauvais:
*Issue #:* /bcgov/entity#6554

*Description of changes:*
- removed border rounding from card (to allow proper error border)
- removed unnecessary row whitespace
- removed RSN hint (to be consistent with other fields)
- updated RSN rule message (to be consistent with other fields)
- fixed validation handler to only validate the current form
- changed reset() to resetValidation() (so input events are not fired)
- updated NONE value for easier comparison (NaN vs -1)
- updated unit test

### DIFF
--- a/src/components/StaffPayment/StaffPayment.vue
+++ b/src/components/StaffPayment/StaffPayment.vue
@@ -1,6 +1,6 @@
 <template>
-  <v-card flat id="staff-payment-container" class="mt-4 py-8 pr-6">
-    <v-row no-gutters class="mt-3">
+  <v-card flat id="staff-payment-container" class="mt-4 py-8 pr-6 rounded-0">
+    <v-row no-gutters>
       <v-col cols="3" class="pl-5" v-if="displaySideLabel">
         <label class="side-label" :class="{'error-text': invalidSection}">Payment</label>
       </v-col>
@@ -11,10 +11,8 @@
           <v-form class="ml-8" ref="fasForm" v-model="fasFormValid">
             <v-text-field
               filled
-              persistent-hint
               id="routing-slip-number-textfield"
               label="Routing Slip Number"
-              hint="FAS Routing Slip Number (9 digits)"
               :value="staffPaymentData.routingSlipNumber"
               :rules="routingSlipNumberRules"
               :disabled="paymentOption === StaffPaymentOptions.BCOL || paymentOption === StaffPaymentOptions.NO_FEE"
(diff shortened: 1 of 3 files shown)


src/components/StaffPayment/StaffPayment.vue
```diff
@@ -172,30 +170,31 @@ export default class StaffPayment extends Vue {
 
   /** Prompt the field validations. */
   @Watch('validate')
-  private validateFields (): void {
-    this.$refs.fasForm.validate()
-    this.$refs.bcolForm.validate()
+  private validateFields (val: boolean): void {
+    if (val) {
+      if (this.paymentOption === StaffPaymentOptions.FAS) this.$refs.fasForm.validate()
+      if (this.paymentOption === StaffPaymentOptions.BCOL) this.$refs.bcolForm.validate()
```
severinbeauvais, line 176: We only want to validate the correct form when validation is enabled. (This should also fix weirdness in the other apps that use this component.)

src/components/StaffPayment/StaffPayment.vue
```diff
@@ -172,30 +170,31 @@ export default class StaffPayment extends Vue {
...
 
   /** Called when payment option (radio group item) has changed. */
   @Watch('paymentOption')
   private onPaymentOptionChanged (val: number): void {
-    this.validateFields()
     switch (val) {
       case StaffPaymentOptions.FAS:
         // reset other form and update data
-        this.$refs.bcolForm.reset()
+        this.$refs.bcolForm.resetValidation()
```
severinbeauvais, line 186: When reset is called, the `@input` event fires, which causes a bunch of events/validation that we don't care about.

Conversely, when resetValidation is called, the validation errors/state are cleared and that's it. Nice and clean.

src/enums/staffPaymentsOptions.ts
```diff
@@ -1,5 +1,5 @@
 export enum StaffPaymentOptions {
-    NONE = NaN,
+    NONE = -1,
```
severinbeauvais, line 2: This is because you can't check equality with NaN. (You have to use isNaN() instead.)

Now we can indeed check `if (val === StaffPaymentOptions.NONE)...`.

src/components/StaffPayment/StaffPayment.vue
```diff
@@ -1,6 +1,6 @@
 <template>
-  <v-card flat id="staff-payment-container" class="mt-4 py-8 pr-6">
-    <v-row no-gutters class="mt-3">
+  <v-card flat id="staff-payment-container" class="mt-4 py-8 pr-6 rounded-0">
```
cameron-freshworks, line 2: oo nice. Is this a border helper?

cameron-freshworks (approved):
👍

src/components/StaffPayment/StaffPayment.vue
severinbeauvais: Indeed. Seek and you shall find. :)

PR #77, opened 2021-06-09
7398 - Replaced folio number input with the shared component

andrepestana-aot:
*Issue #:* /bcgov/entity#7398

*Description of changes:*

- Replaced folio number input with the shared component


src/components/StaffPayment/StaffPayment.vue
```diff
@@ -44,15 +44,14 @@
               @focus="paymentOption = StaffPaymentOptions.BCOL"
               @input="emitStaffPaymentData({ option: StaffPaymentOptions.BCOL, datNumber: $event })"
             />
-            <v-text-field
-              filled
-              id="folio-number-textfield"
-              label="Folio Number (Optional)"
-              :value="staffPaymentData.folioNumber"
-              :rules="folioNumberRules"
+            <FolioNumberInput
```
severinbeauvais, line 47: Since this is a "significant" change to this shared component, do you think we should increase its version to 1.2.0 or 2.0.0 ?
(Is it fully compatible with the previous version?)

andrepestana-aot: I don't think so. It's totally compatible with the previous version.

severinbeauvais (approved):

cameron-freshworks (approved):

PR #78, opened 2021-06-09
7398 - Added validate method for folio number validation

andrepestana-aot:
*Issue #:* /bcgov/entity#7398

*Description of changes:*

- Added validate method for folio number validation to be called by the parent


severinbeauvais (approved):
OK!

PR #84, opened 2021-09-07
8780 - Changed maxNumberOfShares to be number

andrepestana-aot:
*Issue #:* /bcgov/entity#8780

*Description of changes:*

- Changed maxNumberOfShares to be number
- 


src/components/ShareStructure/EditShareStructure.vue
```diff
@@ -281,7 +281,7 @@ export default class EditShareStructure extends Mixins(CurrencyLookupMixin) {
     let rules: Array<Function> = []
     if (!this.hasNoMaximumShares) {
       rules = [
-        (v: string) => !!v || 'Number of shares is required',
+        (v: string) => (v !== '' && v !== null && v !== undefined) || 'Number of shares is required',
```
severinbeauvais, line 284: Should `v` be a number instead of a string here?

Also, should you check for `isNaN()` instead of, or as well as empty, null and undefined?

If there are unit tests for this component, please add some checks for valid and invalid Maximum Number of Shares.

andrepestana-aot: If it is a number it doesn't show the message: 'Number of shares is required'. 

The unit are very well done and captures these conditions. I just made sure the tests are passing with the changes.

severinbeauvais: Thanks for the explanation.

So you declare that the v-model is of type number, but the rules cast v as a string? Weird.

andrepestana-aot: if you don't declare v-model as number it stores it as a string and the IF doesn't validate that. That's why it is required. But we still need to distinguish zero from empty/null/undefined to display different messages to the user.

cameron-freshworks (approved):
Looks good!

severinbeauvais (approved):

PR #92, opened 2021-11-02
9051 Effective Date Time component

severinbeauvais:
*Issue #:* /bcgov/entity#9051

*Description of changes:*
- deleted unnecessary package lock files (revert from previous PR)
- EffectiveDateTime component 
- unit tests
- updated Storybook


src/components/EffectiveDateTime/EffectiveDateTime.vue
```diff
@@ -0,0 +1,477 @@
...
+          </p>
+        </v-col>
+      </v-row>
+    </v-form>
+  </v-card>
+</template>
+
+<script lang="ts">
+import { Component, Emit, Mixins, Prop, Vue, Watch } from 'vue-property-decorator'
+import { DatePicker } from '@/components/DatePicker' // NB: local component
```
severinbeauvais, line 77: Ideally we would import the bcrs-shared-component DatePicker here, but I was getting Storybook errors here (while everything else worked).
I killed a few hours on this so ... this is what I'm going with atm.

src/components/FeeSummary/package-lock.json
```diff
@@ -1,23 +0,0 @@
-{
```
severinbeauvais, line 1: It turns out we probably don't need these package lock files after all.

Oddly, "lerna hoist" created these files once when I ran it. But then I opened up a new window and ran everything from scratch, they weren't created (and everything still worked).

So, I've deleted all these now :)

src/components/index.ts
```diff
@@ -0,0 +1,2 @@
+// local component
+export { DatePicker } from './DatePicker'
```
severinbeauvais, line 2: This is here so that Storybook can find the "local" DatePicker component. The app needs to do the same thing (but there, DatePicker will be an external component).

This is because Storybook can't find `@bcrs-shared-components/DatePicker` ... I don't know why. Lots of time spent on this already :(  This works and it's not terribly ugly.

src/components/EffectiveDateTime/EffectiveDateTime.vue
```diff
@@ -0,0 +1,477 @@
...
+  position: relative;
+
+  .validation-alert-msg {
+    line-height: 12px;
+    position: absolute;
+    top: -2rem;
+    padding: 0 12px;
+    font-size: 12px;
+    font-weight: 500;
+    color: $BCgovInputError !important;
```
cameron-freshworks, line 463: Are these defined in the shared components?

src/components/EffectiveDateTime/EffectiveDateTime.vue
```diff
@@ -0,0 +1,477 @@
...
+      !this.isUnderTime &&
+      !this.isOverTime &&
+      validDateText
+    )
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
```
cameron-freshworks, line 405: Related to the other question, is this something we have defined here? or could this present an issue if pulled in and use in a project without this hierarchy.

severinbeauvais: Yes, if a parent project is missing this folder/file, this will fail. But it has always been this way for all our projects.

All our projects (except NR UI) use the folders above.

cameron-freshworks: Ok cool

src/components/EffectiveDateTime/EffectiveDateTime.vue
severinbeauvais: Yes, these CSS variables are defined in https://github.com/bcgov/bcrs-shared-components/blob/main/src/assets/styles/theme.scss.

And yes, the parent app needs to have these defined as well.

I think/hope the build steps will identify any incorrect imports/variables.

cameron-freshworks (approved):
Looks good!

src/components/EffectiveDateTime/EffectiveDateTime.vue
cameron-freshworks: Copy that!

src/components/EffectiveDateTime/EffectiveDateTime.vue
severinbeauvais: I updated this. See comment on `src/components/index.ts` below.

TLDR: this line is now more obviously a "local import" and it looks the same in the parent app; but the `components/index.ts` file will reference either the local file or the external file, respectively. This keeps things generic/consistent as much as possible while still providing the needed references.

eve-git (approved):
Nice!

PR #93, opened 2021-11-18
8913 Add initialValue prop to DatePicker component

argush3:
*Issue #:* /bcgov/entity#8913

*Description of changes:*

* Added `initialValue` prop to `DatePicker` component.  This was added specifically for the special resolutions section of a Coop voluntary dissolution.  This component is used in draft filings and there is a need to redisplay a selected date when resuming a draft.
* Added unit test to test `initialValue` prop works on `DatePicker` component


tests/unit/DatePicker.spec.ts
```diff
@@ -162,4 +164,13 @@ describe('DatePicker component', () => {
...
   })
+
+  it('displays the correct initialValue', async () => {
+    wrapper = createComponent(null, null, false, [], '2021-11-18')
+    await Vue.nextTick()
+
+    // Verify model and date text field value is updated to use initialValue
+    expect(wrapper.vm.$data.dateText).toEqual('2021-11-18')
+    expect((<HTMLInputElement>wrapper.find('#date-text-field').element).value).toEqual('2021-11-18')
+  })
```
cameron-freshworks, line 175: Nice test!

andrepestana-aot (approved):
LGTM

cameron-freshworks (approved):
Looks good! 
Also, if you wanted to update the `DatePicker.stories` you can sort of mock scenarios in the component regarding this prop. 
Not required, as the interface will allow you to input this prop there either way :)

lekshmimallika-aot (approved):
Looks good!

tests/unit/DatePicker.spec.ts
```diff
@@ -25,14 +25,16 @@ function createComponent (
   title: string = '',
   errorMsg: string = '',
   disablePicker: boolean = false,
-  inputRules = [(v: string) => v === 'Valid text' || 'Invalid']
+  inputRules = [(v: string) => v === 'Valid text' || 'Invalid'],
+  initialValue: string = ''
```
severinbeauvais, line 29: I think this should be `null`, and include a test for this not being set, to verify that existing users of this component don't break.

severinbeauvais: I see `dateText` is initialized to '' in the component, line 91, so the new code shouldn't break anything. (Though, possibly, to avoid multiple initializations, I'd change line line 91 to assign null.)

I'd still like to see a test where `initialValue` is not set, to verify that dateText is initialized to ''.

severinbeauvais: I see `dateText` is initialized to '' in the component, line 91, so the new code shouldn't break anything. Though, possibly, to avoid multiple initializations (*), I'd change line line 91 to assign null.

I'd still like to see a test where `initialValue` is not set, to verify that dateText is initialized to ''.

(*) = multiple initializations is the "decoy pattern": it looks like code is doing something, but it isn't

src/components/DatePicker/DatePicker.vue
```diff
@@ -110,6 +113,9 @@ export default class DatePicker extends Vue {
   @Watch('dateText')
   @Emit('emitDateSync')
   private emitDateSync (date: string): string { return this.dateText }
+
+  @Watch('initialValue', { immediate: true })
```
severinbeauvais, line 117: Since this watcher is only ever used when the component is created (ie, immediately), could you move this assignment right into line 91?

argush3: i'll give it a try.  It makes sense I think.

argush3: doesn't seem like i can do a direct assignment of the prop `initialValue` on line 91.  i imagine vue forces you to do assignment to another variable within watchers or other lifecycle hooks like `created`.  did you want me to do the initial assignment in `created`?

severinbeauvais: Sure, good idea.

argush3: added `created` hook and assignment of `initialValue` to `dateText` is done there now.

severinbeauvais (approved):
Looks good.

Also, I like the new story! 👍

tests/unit/DatePicker.spec.ts
argush3: `dateText` is now initialized as `null` on line 91. 

I added the test where `initialValue` is not set and also updated the prop logic as the base set was a bit hard to work with.  I'm just manually providing the props in each test as opposed to using the default `createComponent` function.

Also added a story test specifically for `initialValue` prop.

PR #94, opened 2021-11-23
8913 Add isDateValid function to datepicker component

argush3:
*Issue #:* /bcgov/entity#8913

*Description of changes:*

* Add public `isDateValid` function to `DatePicker` component.  The validation logic in the special resolutions component for COOP voluntary dissolutions filing had a need to be able to check the validity of the selected date.


src/interfaces/form-interface.ts
```diff
@@ -1,6 +1,6 @@
 import Vue from 'vue'
```
lekshmimallika-aot, line 1: Is this needed?

lekshmimallika-aot (approved):
Looks good

src/components/DatePicker/DatePicker.vue
```diff
@@ -102,6 +104,11 @@ export default class DatePicker extends Vue {
     return this.$refs.form.validate()
   }
 
+  /** Returns whether date validation passes. */
+  public isDateValid (): boolean {
+    return this.$refs.dateTextField.valid
```
severinbeauvais, line 109: You should use conditional chaining in case `dateTextField` is undefined (ie, before the component is initially rendered).

severinbeauvais: Please add a unit test or 2 to verify this this returns True/False as expected.

argush3: will get that updated.

src/interfaces/form-interface.ts
```diff
@@ -1,6 +1,6 @@
...
 
-export interface FormIF extends Vue {
+export interface FormIF extends HTMLFormElement {
```
severinbeauvais, line 3: This is correct. Now we just need to update this in Filings UI (all the others are correct)... or better yet, just import this shared IF :)

Up to you whether to do it as part of your current ticket or create a new ticket.

src/interfaces/form-interface.ts
argush3: good catch.  will remove it.

src/components/DatePicker/DatePicker.vue
argush3: and yes, tests too.  keeping me honest....thx

src/interfaces/form-interface.ts
argush3: removed

src/interfaces/form-interface.ts
severinbeauvais: ^^ I'll update Filings UI in a PR I'm already working on.

src/components/DatePicker/DatePicker.vue
argush3: updated to use conditional chaining and ended up adding a bunch of tests

image

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "esnext",
+    "target": "ES2018",
```
argush3, line 3: turns out storybook 6 does not support conditional chaining. `npm run storybook:build` gave me issues.  i updated to target `ES2018` to get things working.  https://github.com/storybookjs/storybook/issues/11515

severinbeauvais (approved):
Looks good!

PR #95, opened 2021-11-29
8913 DatePicker UX Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#8913

*Description of changes:*
* Updates the display format for DatePicker TextField date
* Functionality to close DatePicker Calendar on route changes
* Updated UX Hover/Disable functionality. 
* Updated Unit testing


src/components/DatePicker/DatePicker.vue
```diff
@@ -19,15 +19,15 @@
...
+                      :value="displayDate"
                       :label="title"
                       :name="Math.random()"
                       :rules="inputRules"
                       :disabled="disablePicker"
-                      v-on="on"
                       v-on:keydown="$event.preventDefault()"
                       v-on:keyup.enter="emitDate(dateText)"
                       v-on:click:append="on.click"
+                      v-on="on"
```
severinbeauvais, line 30: All these v-ons look funny. How about using the shorthand for events?

https://vuejs.org/v2/guide/syntax.html#v-on-Shorthand

src/mixins/date-mixin.ts
```diff
@@ -101,6 +101,28 @@ export default class DateMixin extends Vue {
     return timeStr
   }
 
+  /**
+   * Converts a Date object to a date string (Month Day, Year).
+   * @example "2021-01-01 08:00:00 GMT" -> "Jan 1, 2021"
+   */
+  utcDateToDisplayDate (date: Date): string {
```
severinbeauvais, line 108: This method has been replaced by `dateToPacificDate`.

Just a note that the date mixin HERE is to allow these shared components to work in StoryBook. However, once these components are imported into a project, they will use the date mixin THERE.

So, yes, there are 2 version of the date mixin. I hope we will resolve this some time, but atm this is the situation.

severinbeauvais: I got rid of unhelpful date adjectives like "displayDate", "formattedDate", "prettyDate", "uglyDate", etc, etc 😄 

Now we have the following date formats:
- JS Date (object)
- "yyyyMmDd" 
- "apiDate"
- "pacificDate" (and pacificTime and pacificDateTime)

cameron-freshworks: This is a new method.
Interesting, i did not know that!
I guess we will need to add this to our Create UI as well 🤔

cameron-freshworks: Hmm not sure what to do then. The Datepicker value is `yyyy-mm-dd` and we want to cast that into `Jan 1, 2021` for example just for display puposes. There is no pacific time conversion which why i opted to keep it as utc for simplicity.
In that context, what would be a fitting name?

cameron-freshworks: I think* this format is the preferred format for display purposes... at least according to the latest UXA.

src/components/DatePicker/DatePicker.vue
cameron-freshworks: Will try!

src/components/DatePicker/DatePicker.vue
```diff
@@ -114,6 +115,11 @@ export default class DatePicker extends Vue {
     this.dateText = this.initialValue
   }
 
+  /** The display Date. */
+  private get displayDate (): string {
+    return this.utcDateToDisplayDate(new Date(this.dateText))
```
severinbeauvais, line 120: Due to date-PTSD, I now cringe when I see `new Date(...)`. Example:

```
> new Date('2021-11-29').toUTCString()
> 'Mon, 29 Nov 2021 00:00:00 GMT'

> new Date('2021-11-29').toISOString()
> '2021-11-29T00:00:00.000Z'

> new Date('2021-11-29').toString()
> 'Sun Nov 28 2021 16:00:00 GMT-0800 (Pacific Standard Time)'

> new Date('2021-11-29').toDateString()
> 'Sun Nov 28 2021'
```

It _really_ depends what you do with the resulting Date object. For example, if you just chop off the time portion, you can see that the date (in Pacific tz) is yesterday. (A Date object is always UTC internally, but it may display in "local" timezone as above.)

Passing this to `dateToPacific()` should work fine, since it explicitly converts it to Pacific timezone.

We use `new Date(...)` more than I'd like in our code. I'm sure some of those are incorrect since they intend to convert a "local" time string to a Date object. Unfortunately "local" depends on the browser's timezone and date and time... which could easily not be Pacific, or could be wrong. That is why I wrote `createUtcDate()` -- it explicitly converts from Pacific to a (UTC) Date object.

cameron-freshworks: Ok right on, i will give this a try. 
When i was dealing with Pacific time prior, the date selected in the calendar and the display date were off by the offset. 
Will try what you suggest above!

severinbeauvais: I believe your current code is fine. This was just a (long-winded) FYI.

src/mixins/date-mixin.ts
severinbeauvais: According to the header for both `dateToPacificDate()` and `utcDateToDisplayDate()`, they do the same thing.

I believe I deleted utcDateToDisplayDate in favour of using dateToPacificDate in multiple places already,

src/components/DatePicker/DatePicker.vue
cameron-freshworks: no no i agree, Dates are a thorn. If we can simplify this, definitely will. 
Having a look 👁️

severinbeauvais: Except... call dateToPacificDate instead of utcDateToDisplayDate.

src/mixins/date-mixin.ts
severinbeauvais: > Hmm not sure what to do then. The Datepicker value is yyyy-mm-dd and we want to cast that into Jan 1, 2021 for example just for display puposes. There is no pacific time conversion which why i opted to keep it as utc for simplicity.
> In that context, what would be a fitting name?

I believe what you have in the date picker component is fine (see my latest comments there).

I think you can delete this new method.

severinbeauvais: > I think* this format is the preferred format for display purposes... at least according to the latest UXA.

Yes, I think so too.

src/components/DatePicker/DatePicker.vue
severinbeauvais: I think we have the date conversion methods we need, now.

(Unfortunately, I don't we have a date mixin in any of our UIs that contains all of them, sigh. Future work.)

cameron-freshworks: Was able to do away with the added method and utlize `yyyyMmDdToPacificDate` that currently exists. 
Since we are strictly just converting the format for a display purpose. 
This also does away with creating a new Date object.
Much cleaner!

src/mixins/date-mixin.ts
cameron-freshworks: Updated to use previously existing `yyyyMmDdToPacificDate`, which works great, i missed this method previously i think.

src/components/DatePicker/DatePicker.vue
severinbeauvais: You'll need to add `yyyyMmDdToPacificDate` to the local date mixin in the shared library.

src/components/DatePicker/DatePicker.vue
cameron-freshworks: Updated to short hand events

src/mixins/date-mixin.ts
cameron-freshworks: Updated!

src/components/DatePicker/DatePicker.vue
cameron-freshworks: Updated!

src/components/DatePicker/DatePicker.vue
```diff
@@ -13,23 +13,24 @@
...
-                      :name="Math.random()"
-                      :rules="inputRules"
-                      :disabled="disablePicker"
-                      v-on="on"
-                      v-on:keydown="$event.preventDefault()"
-                      v-on:keyup.enter="emitDate(dateText)"
-                      v-on:click:append="on.click"
-                      filled
-        />
+        <span :class="{'date-text-field-pointer': enableSelector}" v-on="enableSelector && on">
```
severinbeauvais, line 16: Consider doing a tech demo to explain how v-on works. (I'd like to know 😁 )

severinbeauvais (approved):

src/components/DatePicker/DatePicker.vue
cameron-freshworks: Great idea! I'll have to do some more research myself 😆

PR #98, opened 2021-12-17
Shared BreadCrumb

cameron-freshworks:
*Issue #:* /bcgov/entity#10232

*Description of changes:*
* Implements a simple shared BreadCrumb
* Added BreadCrumb interface
* BreadCrumb Stories
* Unit tests


cameron-freshworks:
> _Issue #:_ /[bcgov/entity#10232](https://github.com/bcgov/entity/issues/10232)
> 
> _Description of changes:_
> 
> * Implements Shared BreadCrumb
> * Added BreadCrumb interface
> * BreadCrumb Stories
> 
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the business-filings-ui license (Apache 2.0).

severinbeauvais:
Could you please attach some screen shots of the different variations? (So we don't have to build locally to test ;) )

cameron-freshworks:
> Could you please attach some screen shots of the different variations? (So we don't have to build locally to test ;) )

Totally, can do! I wish i could share it in Storybook as a preview some how...

src/components/BreadCrumb/BreadCrumb.vue
```diff
@@ -0,0 +1,116 @@
...
+  @Prop({ default: () => [] })
+  readonly breadcrumbs: Array<BreadcrumbIF>
+
+  /**
+   *  The url tied to the breadcrumb back button.
+   *  Cross application requires href.
+   *  Returns to previous route when navigating locally.
+   */
+  private get backUrl (): string {
+    const routeIndex = this.breadcrumbs.findIndex(item => item.to?.name === this.$route?.name)
```
severinbeauvais, line 66: What if no match is found?

cameron-freshworks:
Screen Shot 2021-12-17 at 11 17 09 AM
Screen Shot 2021-12-17 at 11 17 03 AM
Screen Shot 2021-12-17 at 11 16 52 AM
Screen Shot 2021-12-17 at 11 16 27 AM
Screen Shot 2021-12-17 at 11 16 20 AM
Screen Shot 2021-12-17 at 11 16 15 AM

src/interfaces/breadcrumb-interface.ts
```diff
@@ -0,0 +1,7 @@
+export interface BreadcrumbIF {
+  text: string
+  disabled?: boolean
+  exact?: boolean
+  to?: any
+  href?: string
```
severinbeauvais, line 6: Why are these optional? Looks like the template code doesn't handle these undefined properties.

src/components/BreadCrumb/BreadCrumb.vue
```diff
@@ -0,0 +1,116 @@
...
+.active-crumb {
+  text-decoration: underline !important;
+  cursor: pointer !important;
+}
+
+::v-deep .v-breadcrumbs .v-breadcrumbs__divider {
+  color: white !important;
+  margin-bottom: 0;
+}
+</style>
```
severinbeauvais, line 116: Please add a unit test for this.

src/components/BreadCrumb/BreadCrumb.vue
cameron-freshworks: Then something is broken :) 
Without the route the user is most likely not in the application any longer.
I suppose it is a requirement to have a `to` on local starting point, so we can count on that.
`to` is optional according to the interface as we will be passing external crumbs which rely on a href, but locally (in our ui's) we will always have/need `to` routes for local navigation. 

So tldr: this check finds our starting point, regardless of how many href crumbs before our local starting point and subsequently tracks any deeper navigation through the `to`. Although we don't have too many implementations of deeper navigation outside the filings ui.

cameron-freshworks:
> Could you please attach some screen shots of the different variations? (So we don't have to build locally to test ;) )

See pics below!

src/interfaces/breadcrumb-interface.ts
cameron-freshworks: Well, a crumb can have either a to or a href, neither are required although 1 is. 
Not sure how to type that.

src/components/BreadCrumb/BreadCrumb.vue
cameron-freshworks: Yea, will do. thinking on what to test here that is meaningful.

src/components/BreadCrumb/BreadCrumb.vue
severinbeauvais: OK. So we leave it up to the user of this component to configure it correctly or else.

Is there any risk of the configuration being messed up under normal running conditions (since I know it's dynamic)?

severinbeauvais:
> > Could you please attach some screen shots of the different variations? (So we don't have to build locally to test ;) )
> 
> See pics below!

Looks amazing!

src/components/BreadCrumb/BreadCrumb.vue
cameron-freshworks: Well, it's no real risk. 
It will simply exist but not do anything. 
For example, if a user does not pass href or to values, the navigation will not go anywhere. 
If they don't pass in any text, there will be crumb on display and no link etc. 

So essentially, no real risk, any persons implementing this can see the example crumb on the prop or refer to the interface. Functionally it just won't do what they want it to do until they pass it a value.

src/interfaces/breadcrumb-interface.ts
severinbeauvais: Not easily/cleanly. I see you just pass those into the breacrumb item, so let's ignore this for now.

Are `disabled` and `exact` used atm?

src/components/BreadCrumb/BreadCrumb.vue
severinbeauvais: I buy that explanation - thanks.

One thing to watch for is that an unhandled exception can cause a method/component to stop running/displaying. This could be serious depending on what gets broken (imagine something in App or TodoList or ...). In this case I think the risk is low: this component may "crash" but I don't think it will take anything down with it, so I think we're good.

src/interfaces/breadcrumb-interface.ts
cameron-freshworks: Not here but i saw it passed in some of the ui's.

src/components/BreadCrumb/BreadCrumb.vue
cameron-freshworks: Ok, thought of some, implementing!

src/interfaces/breadcrumb-interface.ts
severinbeauvais: If there is no requirement to use disabled and exact then I suggest deleting them.

Yes, this would break UIs that think these do something, but on the other hand those UIs will need a bit of refactoring to use this shared component anyway. I think it would be better to not carry forward properties that do nothing.

cameron-freshworks: Ok right on, will do!

lekshmimallika-aot (approved):

src/components/BreadCrumb/BreadCrumb.stories.ts
```diff
@@ -0,0 +1,61 @@
...
+  components: { BreadCrumb },
+  template: '<BreadCrumb v-bind="$props" />' // $props comes from args below
+})
+
+export const homeRoute = Template.bind({})
+homeRoute.args = {
+  breadcrumbs: [
+    {
+      text: 'BC Registries Dashboard',
+      disabled: false
```
severinbeauvais, line 23: Obsolete property. Also below.

tests/unit/BreadCrumb.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+Vue.use(Vuetify)
+Vue.use(Affix)
+
+const vuetify = new Vuetify({})
+
+const homeBreadCrumb = [
+  {
+    text: 'BC Registries Dashboard',
+    disabled: false
```
severinbeauvais, line 16: Obsolete property. Also below.

severinbeauvais (approved):
Looks good. Nice tests.

I pre-approve so you can merge after remaining comments are resolved.

tests/unit/BreadCrumb.spec.ts
cameron-freshworks: thanks, will do!

PR #100, opened 2021-12-22
Updated Shared BreadCrumb

cameron-freshworks:
*Issue #:* /bcgov/entity#10233

*Description of changes:*
* Added disabled prop to allow for manual disabling of routes.


src/components/BreadCrumb/BreadCrumb.vue
```diff
@@ -29,6 +29,7 @@
               exact
               :to="item.to"
               :href="item.href"
+              :disabled="item.disabled"
```
severinbeauvais, line 32: Have you tested `to`, `href` and `disabled` when they are undefined in the item object?

src/interfaces/breadcrumb-interface.ts
```diff
@@ -2,4 +2,5 @@ export interface BreadcrumbIF {
   text: string
   to?: any
   href?: string
+  disabled?: boolean
```
severinbeauvais, line 5: Whoops, this is a property I previously asked you to remove.

What's it for?

severinbeauvais (approved):

src/components/BreadCrumb/BreadCrumb.vue
severinbeauvais, line 32: This could use another unit test, or at least acknowledgement that this new property exists (in unit tests or story).

cameron-freshworks: You bet, it's often undefined and defaulted to false within the Vuetiy Component. Essentially the same as not including the `disabled` prop.

src/interfaces/breadcrumb-interface.ts
cameron-freshworks: Will disable the crumb manually, by passing the value into our vuetify comp prop.
 Since we don't want to control whether a local route or href is used, this allows manually disabling of links.
 Just gives us flexibility. Otherwise end user has no access to the base components prop.

src/components/BreadCrumb/BreadCrumb.vue
severinbeauvais: Oh, right.

I looked it up, and the default value for `to` and `href` is undefined, but the default value for `disabled` is false, so you should fall back to that (eg, `item.disabled || false`).

https://vuetifyjs.com/en/api/v-breadcrumbs-item/#props

cameron-freshworks:
Closing as it's been strongly suggested that we shouldn't be utilizing the `disabled` prop as it is an incorrect solution.

PR #101, opened 2021-12-22
Shared Crumb Style Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#10233

*Description of changes:*
* Updated class binding to explicitly set cursor.


src/components/BreadCrumb/BreadCrumb.vue
```diff
@@ -107,6 +112,10 @@ export default class BreadCrumb extends Vue {
   cursor: pointer !important;
 }
 
+.inactive-crumb {
+  cursor: default !important;
+}
```
severinbeauvais, line 117: 1. I think `!important` can still be overridden by [more specific CSS](https://www.w3schools.com/css/css_specificity.asp). In some cases I have had to use cascading selectors to override something ([example](https://github.com/severinbeauvais/business-filings-ui/blob/8991a1fdabdb37e57bc6205b020d10d1809147df/src/components/dialogs/StaffPaymentDialog.vue#L106)).

2. Would [pointer events](https://www.w3schools.com/cssref/css3_pr_pointer-events.asp) be another, or alternate, property that meets your need?

src/interfaces/breadcrumb-interface.ts
```diff
@@ -2,4 +2,5 @@ export interface BreadcrumbIF {
   text: string
   to?: any
   href?: string
+  disabled?: boolean
```
severinbeauvais, line 5: Obsolete?

cameron-freshworks: removed in next commit

src/components/BreadCrumb/BreadCrumb.vue
cameron-freshworks: 1: In this instance it's working as expected. 
2: I believe it would do the same thing yes.

severinbeauvais: OK, perfect.

src/components/BreadCrumb/BreadCrumb.vue
```diff
@@ -107,6 +112,10 @@ export default class BreadCrumb extends Vue {
   cursor: pointer !important;
 }
 
+.inactive-crumb {
+  cursor: default !important;
```
severinbeauvais, line 116: Is it worth a comment to describe why this is needed?

severinbeauvais (approved):

PR #105, opened 2022-01-17
clear now clears the date

kialj876:
Signed-off-by: Kial Jinnah <kialj876@gmail.com>

*Issue #:* /bcgov/entity#10059

*Description of changes:*
- tiny update to have the clear on the text field clear the date
- test case


severinbeauvais:
Please rebuild Storybook and check those files in.

src/components/DatePicker/DatePicker.vue
```diff
@@ -28,6 +28,7 @@
                         :disabled="disablePicker"
                         :hint="hint"
                         :persistent-hint="persistentHint"
+                        @click:clear="emitCancel()"
```
severinbeauvais, line 31: Do you want this to emit cancel and close the dialog? Or only clear the date?

cameron-freshworks (approved):

src/components/DatePicker/DatePicker.vue
kialj876: > Please rebuild Storybook and check those files in.

Sounds good. What does that do if there isn't a new storybook file? Is it to ensure the existing storybook suites contain the new code?

kialj876:
> Please rebuild Storybook and check those files in.

Sounds good. What does that do if there isn't a new storybook file? Is it to ensure the existing storybook suites contain the new code?

src/components/DatePicker/DatePicker.vue
kialj876: for my case I just want it to emit cancel so that it triggers the same flow

kialj876: otherwise I'll need to add another emit that would be used to trigger the same flow (I thought this was unnecessary)

severinbeauvais:
> What does that do if there isn't a new storybook file? Is it to ensure the existing storybook suites contain the new code?

Yes, since that becomes this (ie, global for the project):

image

> * I updated the readme and took away the comment that says to only do the build for new features

Sounds good.

src/components/DatePicker/DatePicker.vue
argush3: i think we should add a separate emit for the clear.  That way we don't run into any undesired side effects if emitCancel is ever changed.

severinbeauvais: So if the user clicks Cancel then it clear the date? I think that might cause some users to accidentally clear their date (thinking that Cancel would just close the dialog). Could you ask Tracey for her thoughts on this?

severinbeauvais: Note: this isn't emitting a "clear" event. Clearing the dialog is a local function.

severinbeauvais: Oh I get it... this is the "X" on the text field that is used to clear the input.

severinbeauvais: To self-document this (since a couple of us were obviously confused), I recomment adding a new click event function (eg, onClickClear) that, in turn, calls emitCancel.

kialj876: 😢

kialj876: 'It's a one line change it'll be 2 secs' - famous last words

kialj876: If I'm going back in there I will add a new emit fn in case anyone in the future has a separate flow for the cancel vs clear

kialj876: Do we like the current change? New emit + clear ?

argush3: i'm good with this.

argush3 (approved):

severinbeauvais (approved):

PR #106, opened 2022-01-27
Trim payment fields

jordiwes:
*Issue #:* /bcgov/entity#10342

*Description of changes:*
- trim fields in case of accidental white space


src/components/StaffPayment/StaffPayment.vue
```diff
@@ -230,10 +233,10 @@ export default class StaffPayment extends Vue {
   @Emit('update:staffPaymentData')
   private emitStaffPaymentData ({
     option = this.staffPaymentData.option,
-    routingSlipNumber = this.staffPaymentData.routingSlipNumber || '',
-    bcolAccountNumber = this.staffPaymentData.bcolAccountNumber || '',
-    datNumber = this.staffPaymentData.datNumber || '',
-    folioNumber = this.staffPaymentData.folioNumber || '',
+    routingSlipNumber = this.staffPaymentData.routingSlipNumber?.trim() || '',
```
cameron-freshworks, line 236: Does it need to be trimmed here as well if it is handled on the input?
I don't see a problem with this, just curious! 

Does Folio number need trim on the model in the template as well?

src/components/StaffPayment/StaffPayment.vue
```diff
@@ -230,10 +233,10 @@ export default class StaffPayment extends Vue {
...
+    bcolAccountNumber = this.staffPaymentData.bcolAccountNumber?.trim() || '',
+    datNumber = this.staffPaymentData.datNumber?.trim() || '',
+    folioNumber = this.staffPaymentData.folioNumber?.trim() || '',
```
severinbeauvais, line 239: Why do you need to do both? Doesn't `v-model.trim` cause the variable to be trimmed?

Conversely, if the `@input` calls emitStaffData() on every change, that should cause the state variables to update, which the v-text-field should pick up, right?

src/components/StaffPayment/StaffPayment.vue
jordiwes: I have them both in because the vmodel.trim trims the leading spaces in real time (this avoids validation messages like must have 9 characters or whatnot).
What vmodel.trim doesn't seem to do is trim trailing spaces. Even if you tab/click out of the field, which is why I left the manual trim in as well.

src/components/StaffPayment/StaffPayment.vue
jordiwes: see above, I was noticing that just one was not enough ^

src/components/StaffPayment/StaffPayment.vue
cameron-freshworks: Sounds good!

cameron-freshworks (commented):
Sorry for letting this sit! 

Looks fine to me :) 
I would certainly get this branch up to date with main however before rebuilding and pushing up the next commit! I think there was changes to this component specifically as well as several others 👍

severinbeauvais:
@jordiwes Hi Steph.

I think you can just rebuild Storybook and commit that, and it will resolve your merge conflict.

Then Cam or I will merge this and handle the Lerna/publish steps for you. Once that's done, you'll be able to import the latest version of this component.

cameron-freshworks (approved):
Looks good, thanks!

PR #109, opened 2022-02-18
Contact Info - Updated labels and styling

cameron-freshworks:
*Issue #:* /bcgov/entity#10613

*Description of changes:*
* Added label prop to override the contact information descriptor
* updated stories


lekshmimallika-aot (approved):

src/components/ContactInfo/ContactInfo.vue
```diff
@@ -2,9 +2,9 @@
   <div id="contact-info">
     <!-- Display Contact Info -->
     <v-layout row v-if="!isEditing" class="mx-0">
-      <v-flex xs3>
+      <v-flex xs3 class="pr-10">
```
severinbeauvais, line 5: Why add right padding?

For me, right padding is the same as bottom padding, and I try to avoid it.

Also, would it be quick to change this over to using rows and columns (since layout and flex are deprecated)? This might also solve your layout issues.

src/components/ContactInfo/ContactInfo.vue
```diff
@@ -255,9 +252,9 @@ export default class ContactInfo extends Vue {
   @Prop()
   private hasBusinessContactInfoChange!: boolean
 
-  /** Custom contact info msg. */
-  @Prop({ default: null })
-  private customMsg!: string
+  /** Contact information label. */
+  @Prop({ default: 'Registered Office' })
+  private contactLabel!: string
```
severinbeauvais, line 257: So who was using `customMsg` before? They will need to be updated.

Also, how about adding a unit test to verify this prop's output in the template?

src/components/ContactInfo/ContactInfo.vue
cameron-freshworks: I didn't want to expand the size of this column but wanted to limit the number of words on the first row, so to preserve the alignment, this breaks the text after two words. If there is an alternative to limiting the words in the row to 2 words i'm all for it, wasn't sure what else would work here. 

The `<br>` tag won't work for all iterations, for example the design calls for some words in the template on the first line and in other designs the second line etc. 

I don't believe updating the Vue syntax would solve this, i would prefer to defer this with the scope of work on my plate atm.

src/components/ContactInfo/ContactInfo.vue
cameron-freshworks: Good catch, i didn't remove this intentionally. Will reapply this prop!

src/components/ContactInfo/ContactInfo.vue
cameron-freshworks: I could try a character limit in the column vs padding, will try

cameron-freshworks: I was wrong, updating the syntax has helped. Thank you 😁

severinbeauvais (approved):
LGTM. Proceed when you're happy with your changes.

PS unit tests?

cameron-freshworks:
> LGTM. Proceed when you're happy with your changes.
> 
> PS unit tests?

Updated!

PR #111, opened 2022-03-03
Contact Info - Optional Phone Prop

cameron-freshworks:
*Issue #:* /bcgov/entity#10687

*Description of changes:*
* Prop to allow optional phone number rules & label as per evolving requirements.


vysakh-menon-aot (approved):

src/components/ContactInfo/ContactInfo.vue
```diff
@@ -283,6 +283,10 @@ export default class ContactInfo extends Vue {
   @Prop({ default: false })
   private invalidSection!: boolean
 
+  /** Option to disable phone requirement. */
+  @Prop({ default: false })
+  private optionalPhone!: boolean
```
severinbeauvais, line 288: Props should be readonly (and no !).

Oops, please fix the other props, too.

severinbeauvais (approved):
Do you have a new story or test case for this new option?

Also, I wonder if there's a way in GH to delete all previous versions of a file? I'm thinking we don't need to keep any of the Storybook files before the current one.

src/components/ContactInfo/ContactInfo.vue
cameron-freshworks: Copy that, will do!

cameron-freshworks:
> Do you have a new story or test case for this new option?
> 
> Also, I wonder if there's a way in GH to delete all previous versions of a file? I'm thinking we don't need to keep any of the Storybook files before the current one.

For sure, will add another story. 
Being that it's a prop, it can also be toggled in any of the existing stories as well, a nice feature :)

src/components/ContactInfo/ContactInfo.vue
cameron-freshworks: Updated

severinbeauvais: Thanks.

Next time you see a prop with a `!`, it can be removed. (That's a signal to the compiler that the variable might initially be undefined, but that only applies to getters and actions, not props.) (This is not the "non-null assertion operator".)

cameron-freshworks:
> > Also, I wonder if there's a way in GH to delete all previous versions of a file? I'm thinking we don't need to keep any of the Storybook files before the current one.
> 
> I found this: https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#removing-files-from-a-repositorys-history and this: https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#removing-a-file-that-was-added-in-an-earlier-commit
> 
> Next time one of us is working in this repo, we should remove all the old versions of the docs folder.

I didn't even notice this before, thanks! 
Will do, certain i'll be back in here very soon

src/components/ContactInfo/ContactInfo.vue
cameron-freshworks: Ahh i see. 
Considering we also set defaults in a lot of instances. 

Will do 👍

PR #112, opened 2022-03-04
Contact - Minor Layout Adjustments

cameron-freshworks:
*Issue #:* /bcgov/entity#10687

*Description of changes:*
* Updated some layouts to meet spacing requirements


src/components/ContactInfo/ContactInfo.vue
```diff
@@ -186,7 +186,7 @@
             >
             </v-text-field>
           </v-col>
-          <v-col>
+          <v-col class="ml-2">
```
severinbeauvais, line 189: Won't this cause it to not line up with other columns on the page?

cameron-freshworks: This is the split between the two input fields. The cols are on auto, so splitting the margin between the two elements keeps it equal. 

Screen Shot 2022-03-04 at 10 01 02 AM

severinbeauvais (approved):

src/components/ContactInfo/ContactInfo.vue
severinbeauvais: Oh those fields. I thought this was changing the main columns 1 and 2.

cameron-freshworks: Updated the approach to use just padding on the Col 1 and let the auto cols handle the second. 
Works the same, looks the same. Styling in 1 place though 👍

severinbeauvais: When the phone/extension columns wrap, is the second column indented due to the left margin?

severinbeauvais: Oh nvm -- I see you latest code has only right margin on the first column, so the second column should be fine.

PR #113, opened 2022-03-09
Fee Summary Updates

cameron-freshworks:
*Issue #:* /bcgov/entity#11324

*Description of changes:*
* Updated Fee Summary Layout and Operation
* Updated Stories
* Update Unit Testing


vysakh-menon-aot (approved):

src/components/FeeSummary/FeeSummary.stories.ts
```diff
@@ -21,12 +21,29 @@ const filingData: FilingDataIF = {
...
+  filingTypeCode: FilingCodes.CHANGE_OF_REGISTRATION,
+  entityType: CorpTypeCd.SOLE_PROP,
+  priority: false
+}
+
 export const Default = Template.bind({})
 Default.args = {
   filingData: filingData,
-  payApiUrl: 'https://pay-api-dev.pathfinder.gov.bc.ca/api/v1/',
+  payApiUrl: 'https://pay-api-dev.apps.silver.devops.gov.bc.ca/api/v1/',
```
severinbeauvais, line 33: Nice catch.

Are there any other URLs in shared components that should be updated?

cameron-freshworks: It doesn't appear so, nothing related to the migration to the silver cluster at least. 
The only other urls i could find related to github references.

cameron-freshworks: O, one in a unit test.. although it's relevance is questionable 😁 
Updated

src/components/FeeSummary/FeeSummary.vue
```diff
@@ -5,36 +5,46 @@
       :payURL="payApiUrl"
     />
 
-    <v-row class="mt-1">
-      <v-col cols="8" class="pr-2">
+    <v-row no-gutters>
```
severinbeauvais, line 8: This looks good.

FYI, I've been going the "no gutters" route as well in Create UI. Typical template:
```
<v-row no-gutters>
  <v-col cols="12" sm="3" class="pr-4">
     My Label
  </v-col>

  <v-col cols="12" sm="9">
    My Data
  </v-col>
</v-row>
```

The `cols="12"` is to make the column full-width under size "sm".

`sm` seems to be a good breakpoint for the Create UI entry and review pages.

The `class="pr-4"` is to add a "gutter" to the label column so the label doesn't run into the data column. (The data column does need this.)

src/components/FeeSummary/FeeSummary.vue
```diff
@@ -5,36 +5,46 @@
...
           large
           :loading="isLoading"
-          @click="emitAction(SummaryActions.SAVE_RESUME_LATER)"
+          @click="emitAction(SummaryActions.BACK)"
         >
-          <span>Save and Resume Later</span>
+          <span><v-icon>mdi-chevron-left</v-icon>Back</span>
+        </v-btn>
+      </v-col>
+      <v-col class="pt-3">
```
severinbeauvais, line 19: This is suspicious. Why do you need to add top padding to both columns? This should be done at the row element.

On the other hand, I've found it necessary to add "mt-n1" for some buttons to line up with the other columns.

src/components/FeeSummary/FeeSummary.vue
```diff
@@ -83,6 +93,10 @@ export default class FeeSummary extends Vue {
   @Prop({ default: '' })
   readonly errorMessage: string
 
+  /** Prop to indicate review mode. */
+  @Prop({ default: false })
+  readonly isSummaryMode: boolean
```
severinbeauvais, line 98: Why not call it `isReviewMode` then?

src/components/FeeSummary/FeeSummary.vue
cameron-freshworks: Very cool! 
Yes i definitely agree and try to follow this route whenever possible. 
This component was pretty easy.

src/components/FeeSummary/FeeSummary.vue
cameron-freshworks: Will update comment, changed it summary after the fact to match the precedence set in the UI's

src/enums/filingCodes.ts
```diff
@@ -3,5 +3,6 @@ export enum FilingCodes {
   INCORPORATION_CP = 'OTINC',
   INCORPORATION_BC = 'BCINC',
   CORRECTION = 'CRCTN',
-  ALTERATION = 'ALTER'
+  ALTERATION = 'ALTER',
+  CHANGE_OF_REGISTRATION = 'FMCHANGE'
 }
```
severinbeauvais, line 8: This is a shared enum, right?

Could you please add the missing code? Eg,
https://github.com/bcgov/business-create-ui/blob/main/src/enums/filingCodes.ts

src/components/FeeSummary/FeeSummary.vue
cameron-freshworks: This is to space the buttons vertically without having to write a row for every single button. 
Although it could be done our current requirements and implementation make it pretty easy to just let the vuetify grid handle this, and we apply these padding adjustments. 

Screen Shot 2022-03-09 at 1 58 40 PM

src/enums/summaryActions.ts
```diff
@@ -1,5 +1,6 @@
 export enum SummaryActions {
-  SAVE_RESUME_LATER,
-  DELETE_ALL,
-  CONFIRM
+  BACK,
+  CANCEL,
+  CONFIRM,
+  SAVE_RESUME_LATER
 }
```
severinbeauvais, line 6: I suggest this be renamed to FeeSummaryActions, so it's not confused with our Actions component.

src/components/FeeSummary/FeeSummary.vue
cameron-freshworks: Also allows us to throw an additional button in and it handles the space between.

Screen Shot 2022-03-09 at 1 58 37 PM

src/components/FeeSummary/FeeSummary.vue
cameron-freshworks: Updated!

src/components/FeeSummary/FeeSummary.vue
severinbeauvais: Good answers :)

src/enums/filingCodes.ts
cameron-freshworks: For sure

src/enums/summaryActions.ts
cameron-freshworks: K
Updated

severinbeauvais (approved):
LGTM

PR #114, opened 2022-03-10
11474 Nature Of Business component

severinbeauvais:
*Issue #:* /bcgov/entity#11474

*Description of changes:*
- implemented Nature Of Business component
- updated some interfaces
- added some theme variables
- some lint fixes
- updated component's imports to use its own dependencies (might not work)


severinbeauvais:
Storybook:

image

src/components/NatureOfBusiness/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@bcrs-shared-components/nature-of-business",
+  "version": "0.0.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@bcrs-shared-components/interfaces": "^1.0.22",
```
severinbeauvais, line 8: I'm hoping Lerna will automatically update this import to v23 (which is the version with the new interfaces). We'll see.

severinbeauvais: Maybe it doesn't matter, because this component expects to find the interfaces in "@/interfaces", which exists here and will exist in the parent web app.
So I dunno what this import is for, then. Perhaps to make this component more standalone (ie, don't depend on parent to provide the interfaces)? 🤷‍♂️

src/components/NatureOfBusiness/NaicsResult.vue
```diff
@@ -0,0 +1,97 @@
...
+          Show more...
+        </div>
+      </template>
+    </v-col>
+  </v-row>
+</template>
+
+<script lang="ts">
+import { Component, Emit, Prop, Vue } from 'vue-property-decorator'
+import { NaicsElementIF, NaicsResultIF } from '@bcrs-shared-components/interfaces'
```
severinbeauvais, line 35: This _should_ work. We'll see.
(Same in NaicsResults.vue.)

src/components/NatureOfBusiness/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@bcrs-shared-components/nature-of-business",
+  "version": "0.0.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@bcrs-shared-components/interfaces": "^1.0.31",
```
severinbeauvais, line 8: I'm hoping this version of the dependency contains my new interfaces, or that Lerna will auto-update it. We'll see.

(Locally, ~/src/components/NatureOfBusiness/node_modules/@bcrs-shared-components/interfaces/ contains my new interfaces, and I didn't copy them there. 🤷‍♂️ )

src/components/NatureOfBusiness/NaicsResult.vue
severinbeauvais: Oh wait... "@bcrs-shared-components" might point to the main node_modules, not my component's node_modules. But the unit tests and Storybook worked fine. Hmm...

cameron-freshworks: Not sure either, worth a try tho!

src/components/NatureOfBusiness/package.json
cameron-freshworks: Is this the future version or what it is now? 
I imagine you could put what it is now, and Lerna will update this? 
I could be wrong on this tho..

cameron-freshworks (approved):
Looks good!

src/components/NatureOfBusiness/package.json
severinbeauvais: This is the now version.

PR #116, opened 2022-03-12
11474 Updated layouts and styles in some components

severinbeauvais:
*Issue #:* /bcgov/entity#11474

*Description of changes:*

**THIS IS A MINOR BREAKING CHANGE, since these components not longer render their own "surrounding" v-card; they are now more pure/atomic (div only). The parent page that uses these components will need to wrap these in a v-card / section / sheet / whatever else the parent wants to do to display this component consistently with its other components.**

- updated layouts (rows/cols) and misc styles in Certify component
- updated layouts (rows/cols) and misc styles in ContactInfo component
- updated layouts (rows/cols) and misc styles in StaffPayment component
- changed Certify from v-card to div
- changed StaffPayment from v-card to div
- changed CourtOrderPoa from v-card to div + updated layout
- updated unit tests
- updated Storyboook


src/components/Certify/Certify.vue
```diff
@@ -1,80 +1,78 @@
...
-                  , certify that I have relevant knowledge of the
-                  {{ entityDisplay || "association" }} and I am authorized to
-                  make this filing.
-                </div>
-              </template>
-            </v-checkbox>
-            <ul class="certify-statements mt-4">
-              <li v-for="(statement, index) in statements" :key="`statement-${index}`" class="pt-2">
-                {{ statement }}
+  <v-card flat id="AR-step-4-container" class="py-8 px-6">
```
severinbeauvais, line 2: I'm actually having second thoughts about this... The parent should declare the whitespace (padding) around this v-card, not the component. This will allow the parent to define the style around this component however it chooses. Maybe this shouldn't be a v-card at all, so that the parent can plunk this component into whatever type of element it wants. What do you think?

src/components/Certify/Certify.vue
```diff
@@ -1,80 +1,78 @@
...
+    <v-form ref="certifyForm" lazy-validation v-on:submit.prevent>
+      <v-row no-gutters>
+        <v-col cols="12" :sm="firstColumn" class="pr-4 pb-4">
```
severinbeauvais, line 5: `cols="12"` means this column stacks (becomes full-width) below size sm.
`pr-4` is used to provide a small gutter so the column text doesn't bump into the next column.
`pb-4`is used to provide some whitespace between stacked columns.

src/components/Certify/Certify.vue
```diff
@@ -1,80 +1,78 @@
...
+          <label class="title-label" :class="{'error-text': invalidSection}"><strong>Legal Name</strong></label>
```
severinbeauvais, line 6: We should declare global classes for this kind of label (eg, font-size, font-weight, color). Maybe later.

src/components/ContactInfo/ContactInfo.vue
```diff
@@ -1,120 +1,118 @@
...
+          <div v-if="!!businessContact.phone" id="lbl-phone">
+            {{ businessContact.phone }}
+            <span v-if="!!businessContact.extension">Ext: {{ businessContact.extension }}</span>
+          </div>
+          <div v-else id="lbl-no-phone">(Not entered)</div>
+        </v-col>
+
+        <!-- Edit Actions -->
+        <v-col v-if="!disableActions" cols="6" sm="2">
+          <div class="d-flex justify-end align-end align-sm-start">
```
severinbeauvais, line 38: Flex container for the button, which aligns it to the bottom by default, and the top from size sm and greater.

src/components/ContactInfo/ContactInfo.vue
```diff
@@ -1,120 +1,118 @@
...
-                </v-list-item>
-              </v-list>
-            </v-menu>
-          </span>
-        </div>
-      </v-col>
-    </v-row>
-
-    <!-- Edit Contact Info -->
-    <template v-else id="edit-contact-form">
```
severinbeauvais, line 107: IDs don't work on templates. (Classes don't work either.)

src/components/ContactInfo/ContactInfo.vue
```diff
@@ -201,27 +203,14 @@
...
-                id="cancel-btn"
-                large outlined color="primary"
-                @click="cancelEdit()"
-              >
-                <span>Cancel</span>
-              </v-btn>
-            </div>
-          </v-col>
-        </v-row>
+        <div class="action-btns d-flex justify-center justify-sm-end">
```
severinbeauvais, line 206: Flex container for the buttons, which centers them by default, and to the right side from size sm and greater.

src/components/ContactInfo/ContactInfo.vue
```diff
@@ -377,41 +362,40 @@ export default class ContactInfo extends Vue {
...
   }
+}
 
-  #done-btn[disabled] {
-    color: white !important;
-    background-color: $app-blue !important;
-    opacity: 0.2;
-  }
+#contact-info-undo-btn {
+  border-right: 1px solid $gray1;
```
severinbeauvais, line 392: This is not new. It's from old lines 393-395.

src/components/StaffPayment/StaffPayment.vue
```diff
@@ -1,15 +1,15 @@
 <template>
-  <v-card flat id="staff-payment-container" class="mt-4 py-8 pr-6 rounded-0">
+  <v-card flat id="staff-payment-container" class="py-8 px-6">
```
severinbeauvais, line 2: Again, I'm thinking this shouldn't be a v-card. The parent can wrap this in a v-card if it needs to.

src/components/Certify/Certify.vue
severinbeauvais: I changed it.

src/components/StaffPayment/StaffPayment.vue
severinbeauvais: I changed it.

severinbeauvais:
> Were some requirements updated for these components? Do they still align in there current implementations in the Edit UI as well?

These were implicit requirements: as part of layout cleanup in Create UI for registrations, I took this opportunity to remove the wrappers (v-cards in this case) from these components. The v-cards have styling that should not be baked in to these shared components; they should be more pure -- and leave wrapper styling to the parent page. You will see in my forthcoming Create UI PR that the parent pages no longer reach deep (::v-deep) into various components to override their styles; the pages now simply style the components consistently with the other components on their page. (Also these components are now responsive.)

No, you cannot pull these into Edit UI directly. I will update the release numbers for these components to indicate a "major" change. Edit UI can continue to use the old components, or even better: import the new components and wrap them with a v-card or whatever is needed. These are layout changes only (no functional changes).

PR #121, opened 2022-03-22
11519 Fixed ContactInfo

severinbeauvais:
*Issue #:* /bcgov/entity#11519

*Description of changes:*

- fixed missed events

### DIFF
--- a/src/components/ContactInfo/ContactInfo.vue
+++ b/src/components/ContactInfo/ContactInfo.vue
@@ -281,7 +281,7 @@ export default class ContactInfo extends Vue {
 
   // Local Properties
   private isEditing: boolean = false
-  private contactInfo: ContactPointIF
+  private contactInfo: ContactPointIF = null
   private formValid: boolean = false
 
   // Text-field Rules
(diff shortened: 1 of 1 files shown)


src/components/ContactInfo/ContactInfo.vue
```diff
@@ -341,9 +341,9 @@ export default class ContactInfo extends Vue {
         (v: any) => !v || (v.length === 0 || v.length === 14) || 'Phone number is invalid'
       ]
 
-    // Await the applied rules and validate form
-    await Vue.nextTick()
+    // Validate form and wait for v-model to get updated
     this.$refs.editContactForm.validate()
+    await Vue.nextTick()
```
severinbeauvais, line 346: There is no need to wait to the rules to apply -- `validate()` handles that.

However, we do need to wait for prop `formValid` to get updated by validate.

Another way to do this would be to check the response from validate -- if True then the form is valid, else not.

severinbeauvais: This bug caused the code to see the PREVIOUS value of `formValid`. In practise this means that the first time this method (submitContact) is run, the form is valid, but on subsequent calls it might be invalid, so updateContactInfo() is not called (line 349). The user has to click the OK button a second time.

cameron-freshworks (approved):
Looks good!

PR #125, opened 2022-04-20
Shared Completing Party Component

cameron-freshworks:
*Issue #:* /bcgov/entity#11742

*Description of changes:*
* Implements Shared Component for Completing Party edit and view


severinbeauvais (approved):
Looks good!

cameron-freshworks:
> Looks good!

Thank you! 
Still looking at how to better support the Story for these Common Components who have a multitude of dependencies.

severinbeauvais:
> Still looking at how to better support the Story for these Common Components who have a multitude of dependencies.

Can the dependencies be eliminated? Passed on as parameters instead of implicit?

Should this component live in sbc-common-components due to its more complex nature and dependencies?

cameron-freshworks:
> > Still looking at how to better support the Story for these Common Components who have a multitude of dependencies.
> 
> Can the dependencies be eliminated? Passed in as parameters instead of implicit?
> 
> Should this component live in sbc-common-components due to its more complex nature and dependencies?

I think i was overthinking it (yes, definitely too much thinking). 
The storybook error wasn't based on the common components means of fetching data but on the sub (Vue) components it was using. I needed to give story a vuetify instance to work with. For whatever reason, the components will mount/function properly at the root level, but without vuetify being passed into the story, the sub components would take issue. 
I wish i had more answers lol. Just my findings so far :)

src/assets/styles/layout.scss
```diff
@@ -52,6 +52,10 @@
   }
 }
 
+.section-container {
+  padding: 1.25rem 1.875rem;
+}
```
severinbeauvais, line 57: Fortunately this SCSS file is only used while the components are running locally (Storybook). When imported, the components use the parent's (project's) assets.

Otherwise I'd worry this is changing other components unexpectedly.

src/components/CompletingParty/CompletingParty.stories.ts
```diff
@@ -0,0 +1,73 @@
...
+  deliveryInstructions: {
+    maxLength: maxLength(80)
+  }
+}
+
+const Template = (args, { argTypes }) => ({
+  vuetify: new Vuetify({ iconfont: 'mdi' }),
+  props: Object.keys(argTypes),
+  components: { CompletingParty },
+  template: '<v-card flat ><completingParty v-bind="$props" /></v-card>'
```
severinbeauvais, line 43: I know Vue is "flexible" with component naming but I think this should be `completing-party` or `CompletingParty`.

cameron-freshworks: oops, for sure, will fix to `CompletingParty`

src/assets/styles/layout.scss
cameron-freshworks: Definitely! Why i wasn't too worried about adding it in for the stories sake.

src/components/CompletingParty/CompletingParty.vue
```diff
@@ -0,0 +1,287 @@
...
+              <label class="sub-header">Mailing Address</label>
+              <div class="address-wrapper pt-6">
+                <base-address
+                  ref="mailingAddress"
+                  :editing="true"
+                  :schema="addressSchema"
+                  :address="currentCompletingParty && currentCompletingParty.mailingAddress"
+                  @update:address="completingParty.mailingAddress = $event"
+                  @valid="addressValid = $event"
+                />
```
severinbeauvais, line 65: I think you missed my dev demo yesterday but I talked about an invalid Base Address component not making the form invalid, and I did this to work around it:

https://github.com/bcgov/business-create-ui/blob/a3c99025d0ceb6ad4831ef88c065e80edfa8e3d6/src/components/Registration/RegAddEditOrgPerson.vue#L232

src/components/CompletingParty/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@bcrs-shared-components/completingParty",
+  "version": "0.0.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@bcrs-shared-components/interfaces": "1.0.33",
+    "vue-property-decorator": "^8.4.0"
+  }
```
severinbeauvais, line 10: Is lodash a dependency?

severinbeauvais: Also vuelidate.

src/components/CompletingParty/CompletingParty.vue
severinbeauvais: In one case I had to do both: once everything is valid, go back and edit an address field to make the address form invalid. Does the main form become invalid too?

If you just moved the current code here from a UI, don't worry too much about it. (Kaine has said that he'd rather get more stuff implemented than get small stuff perfect, as long as we don't accept bad data.)

Does this component (and address) validate correctly currently? If so, no change needed here.

src/components/CompletingParty/package.json
severinbeauvais: The root dependencies are for this project as a whole, ie, Storybook.

The local dependencies call out what this particular component needs, which is probably already being provided by the parent project but I think it's still nice to call it out to make this component self-contained / independent of external hidden dependencies.

src/components/CompletingParty/CompletingParty.vue
cameron-freshworks: Isolated yes, when the validate prop is passed it will validate both the Name inputs and the Address inputs against the schema provided. 
Obviously, this state needs to be handled at an app level as well before filing (ie ensure there is a completing party before filing) but that can be handled locally.

PR #130, opened 2022-05-16
4781 Created date string types

severinbeauvais:
*Issue #:* /bcgov/entity#4781

*Description of changes:*
- created date types to differentiate string formats
- updated date mixin
- updated StoryBook


src/interfaces/date-interfaces.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * ISO date string (YYYY-MM-DD), assumed to be in Pacific timezone.
+ * @example "2021-12-31"
+ */
+ export type IsoDatePacific = string
```
severinbeauvais, line 5: This is so we can declare a variable of this type and know what format it should be.

cameron-freshworks (approved):
Looks good!

PR #131, opened 2022-05-26
12310 Nature Of Business tweaks

severinbeauvais:
*Issue #:* /bcgov/entity#12310

*Description of changes:*
- updated code-desc display
- updated state conditions
- updated button right alignment
- updated indents (lint)


src/components/nature-of-business/NatureOfBusiness.vue
```diff
@@ -180,24 +184,16 @@ export default class NatureOfBusiness extends Vue {
 
   /** Whether we have stored NAICS data. */
   get haveNaics (): boolean {
-    return (!!this.naicsCode && !!this.naicsDescription)
+    return (!!this.naicsCode || !!this.naicsDescription)
```
severinbeauvais, line 187: This is to handle the situation where we have a code and no description (or vice-versa).

(Yes, this is now a valid scenario.)

src/components/nature-of-business/NatureOfBusiness.vue
```diff
@@ -180,24 +184,16 @@ export default class NatureOfBusiness extends Vue {
...
     return (this.haveNaics && this.state === States.SUMMARY)
   }
 
-  /** Called when this component has been created. */
-  created (): void {
-    // on init, if we have stored NAICS data then display summary
-    if (this.haveNaics) {
-      this.state = States.SUMMARY
-    }
-  }
```
severinbeauvais, line 197: The problem with this is the state is only updated when the component is created -- not when the props change.

Now we use a watcher instead (see below).

src/components/nature-of-business/NatureOfBusiness.vue
```diff
@@ -285,7 +287,12 @@ export default class NatureOfBusiness extends Vue {
   padding: 8px;
 }
 
+#nob-change-btn {
+  margin-right: -14px;
```
severinbeauvais, line 291: This is so the button label is in line with the right margin, but when highlight the padding is the same on the left and right (it goes past the margin), ie:

image

cameron-freshworks: I think* the Edit UI might already have some compensation for this, so might need an update there, super minor 👍

severinbeauvais: It does. I'm in there already so I removed the v-deep.

PR #132, opened 2022-05-30
Bug/12384 - fix pacific date format bug on date picker display [Draft]

ritvick:
*Issue #:* /bcgov/entity###
https://github.com/bcgov/entity/issues/12384

*Description of changes:*

### DIFF
--- a/src/components/date-picker/DatePicker.vue
+++ b/src/components/date-picker/DatePicker.vue
@@ -131,7 +131,9 @@ export default class DatePicker extends Mixins(DateMixin) {
 
   /** The display Date. */
   private get displayDate (): string {
-    return this.yyyyMmDdToPacificDate(this.dateText, true)
+    if(!this.dateText) return ''
+    const dateStr = new Date(this.dateText)
+    return dateStr.toDateString().split(' ').slice(1).join(' ')
   }
 
   /** True when the picker is not displayed or disabled. */


src/components/date-picker/DatePicker.vue
```diff
@@ -131,7 +131,9 @@ export default class DatePicker extends Mixins(DateMixin) {
 
   /** The display Date. */
   private get displayDate (): string {
-    return this.yyyyMmDdToPacificDate(this.dateText, true)
+    if(!this.dateText) return ''
+    const dateStr = new Date(this.dateText)
+    return dateStr.toDateString().split(' ').slice(1).join(' ')
```
severinbeauvais, line 136: Don't do this. It will change the display date format for all apps that import this shared component. The issue reported in ticket 12384 is likely due to how the app initializes the date field used in this component.

Use the date mixins. They were implemented to handle the various formats we deal with. For example, you should not use `new Date(x)`... there is almost certainly a mixin function to convert from your expected source format to a JS Date object.

src/components/date-picker/DatePicker.vue
```diff
@@ -131,7 +131,9 @@ export default class DatePicker extends Mixins(DateMixin) {
...
   }
```
severinbeauvais, line 137: If you end up changing this code then please add unit tests to verify your change (the misc scenarios).

severinbeauvais (changes requested):

src/components/date-picker/DatePicker.vue
ritvick: Hi @severinbeauvais 
I see what you are saying here. But apart from format, the value of date is itself wrong. 
For example even if for a BC benefit company if I select a dissolution date say 6th June? The datepicker gives 6th June -1 day = 5th June, as the date picker is sending pacific date not the local.

severinbeauvais: Are you saying that the yyyymmdd returned by the Vuetify datepicker is not correctly converted to a Pacific date string here? Please provide examples of what the v-date-picker actually returns here.

ritvick: the date picker is giving a string which is converted to pacific date string. 
The bug https://github.com/bcgov/entity/issues/12384 here reports this should not be expected behaviour. 
If you check the screen shot in the link https://github.com/bcgov/entity/issues/12384, selecting 26th may as date, the textbox gets 25th may as date. 

And this is consistant across all date pickers.

severinbeauvais: Thanks. I will look into this later today.

severinbeauvais: I can't reproduce this for start date in a registration (same example as ticket 12384). I clicked on May 31 and the text box says "May 31, 2022". What did you do to see this?

severinbeauvais: As far as I know, the v-date-picker emits a string. So if you click 31 in May 2022 then it spits out "2022-05-31".

Then this shared component simply emits that string (still 2022-05-31).

What your component does next with that string can vary. If you handle it as a "Pacific date" then it's still May 31, 2022. But if you convert it to a date using `new Date()` then it will use the browser's locale and date/time to convert it, so then you get a UTC object (with whatever standard/daylight offset your browser uses).

This is why the date mixin methods exist, and why knowing the "type" of date you're working with is important, and why you shouldn't use `new Date()`.

If I am not understanding something, please let me know.

cameron-freshworks: If it's related to UTC vs local time, would it require to be set before 7am in the morning to show the previous day?

Is it using a server dateTime or the users local system time to initialize? Perhaps reproducible by changing your system time?

severinbeauvais: Maybe the v-date-picker [locale prop](https://vuetifyjs.com/en/api/v-date-picker/#props) needs to be set explicitly? By default, it's undefined. I _assume_ that means "what you see is what you get" but assumptions are sometimes wrong.

severinbeauvais: > If it's related to UTC vs local time, would it require to be set before 7am in the morning to show the previous day?
> 
> Is it using a server dateTime or the users local system time to initialize? Perhaps reproducible by changing your system time?

I assume/hope the v-date-picker is timezone-agnostic = what you see is what you get (May 31, 2022 is always emitted as 2022-05-31).

ritvick: Yeah this is an issue for the users which has timezone other than pacific. 
If you set your timezone to EST, you will see the behaviour mentioned the the bug ticket.

severinbeauvais: > Yeah this is an issue for the users which has timezone other than pacific. If you set your timezone to EST, you will see the behaviour mentioned in the bug ticket.

Please add this to the ticket description :)

severinbeauvais: After changing my PC to EST and 3am, I have now reproduced this bug.

After selecting "May 31, 2022", these are the values in the shared date picker component:

image

I believe the bug is in date mixin method `yyyyMmDdToPacificDate()` (or one of the methods it calls). It was provided a date string of "2022-05-31" and it returned string, "May 30, 2022".

ritvick:
The fix for this was done in app date mixin:
https://github.com/bcgov/business-create-ui/pull/402

PR #135, opened 2022-07-12
12872 Fixed id mismatches between string/number

severinbeauvais:
*Issue #:* /bcgov/entity#12872

*Description of changes:*
- fixed id mismatches between string/number
- updated Storybook


src/components/share-structure/ShareStructure.vue
```diff
@@ -795,14 +795,14 @@ export default class ShareStructure extends Vue {
     // Fetch the original Share class ( In the event the list is moved up or down, find the original by ID )
     const originalShareClass = Object.assign({},
       originalShareClasses.find(
-        shareClass => shareClass.id === parentId
+        shareClass => (+shareClass.id === +parentId)
```
severinbeauvais, line 798: In some cases, id is a string. In other case, id is a number. Triple-equal matches the type and value, so these `find` calls often failed to return any results, which broke the Undo functionality.

lewischenstudio (approved):

src/components/share-structure/ShareStructure.vue
lewischenstudio: This unary plus operator in JS works like a charm!

PR #144, opened 2022-08-19
13399 Certify component not displaying red error bar and texts

lewischenstudio:
*Issue #:* /bcgov/entity#13399

*Description of changes:*
- fixed and enabled "invalid-section" for the certify component
- enforced css
- added unit test

### DIFF
--- a/src/components/certify/Certify.vue
+++ b/src/components/certify/Certify.vue
@@ -192,6 +192,12 @@ export default class Certify extends Vue {
   line-height: 1.2rem;
   font-size: $px-16;
 
+  &.invalid-section {
+    border-left: 3px solid $app-red !important;
+    border-top-left-radius: 0 !important;
+    border-bottom-left-radius: 0 !important;
+  }
+
   .error-text {
     color: $app-red;
   }
--- a/src/components/certify/Certify.vue
+++ b/src/components/certify/Certify.vue
@@ -1,5 +1,5 @@
 <template>
-  <div id="AR-step-4-container">
+  <div id="AR-step-4-container" :class="{'invalid-section': invalidSection}">
     <v-form ref="certifyForm" lazy-validation v-on:submit.prevent>
       <v-row no-gutters>
         <v-col cols="12" :sm="firstColumn" class="pr-4 pb-4">
@@ -191,6 +191,10 @@ export default class Certify extends Vue {
 #AR-step-4-container {
   line-height: 1.2rem;
   font-size: $px-16;
+
+  .error-text {
+    color: $app-red;
+  }
 }
 
 .title-label {
--- a/src/components/certify/Certify.stories.ts
+++ b/src/components/certify/Certify.stories.ts
@@ -22,5 +22,6 @@ Default.args = {
   message: 'Note: It is an offence to make a false or misleading statement in respect of a material fact in a \n' +
     'record submitted to the Corporate Registry for filing. See section 427 of the Business Corporations Act.',
   entityDisplay: 'BC Company',
-  disableEdit: false
+  disableEdit: false,
+  invalidSection: false
 }
--- a/tests/unit/Certify.spec.ts
+++ b/tests/unit/Certify.spec.ts
@@ -138,6 +138,7 @@ describe('Certify', () => {
     const wrapper: Wrapper<Certify> =
       createComponent(null, true, false, null, false, true)
 
+    expect(wrapper.find('.invalid-section').exists()).toBe(true)
     expect(wrapper.find('.error-text').exists()).toBe(true)
   })
 
--- a/tests/unit/date-mixin.spec.ts
+++ b/tests/unit/date-mixin.spec.ts
@@ -76,3 +76,4 @@ describe('Date Mixin', () => {
     expect(vm.apiToPacificDateTime('2021-01-01T00:00:00+00:00')).toBe('December 31, 2020 at 4:00 pm Pacific time') // PST
     expect(vm.apiToPacificDateTime('2021-07-01T00:00:00+00:00')).toBe('June 30, 2021 at 5:00 pm Pacific time') // PDT
   })
+})


lewischenstudio:
Steps I did on my local environment:
- I ran storybook to verify the "invalidSection" variable is working properly.
- I used Node 12 and ran `npm i` successfully without the `package-lock.json` file
- I ran `npm run storybook:build` which updated too many files inside `doc` folder, which wasn't committed.

image

image

src/components/certify/Certify.vue
```diff
@@ -191,6 +191,16 @@ export default class Certify extends Vue {
...
+
+  &.invalid-section {
+    border-left: 3px solid $app-red !important;
+    border-top-left-radius: 0 !important;
+    border-bottom-left-radius: 0 !important;
+  }
+
+  .error-text {
+    color: $app-red;
+  }
```
severinbeauvais, line 203: Aren't both of these classes already present in the base SCSS, both here in bcrs-shared-components and in the UI projects where this component is imported?

lewischenstudio: Yes, they are. I just wanted to make sure they do. Give me a minute. Let me check if they are persisted on storybook. I'll remove it if they are.

lewischenstudio: image

I removed it and checked storybook. Somehow they weren't there anymore. That's why it isn't shown on Dev either.

Dev below:

image

lewischenstudio: I think they were not persisted in this component, so this code just fixed it.

severinbeauvais: Dev and Storybook aren't the same thing. Storybook is a project in THIS repo that loads this component. Dev could be Filings/Create/Edit/other UI that uses this component.

Please add these classes to the bcrs-shared-components assets.

Also, if you identify a UI that doesn't display these classes, add the classes to that project.

We don't want the duplicate class declarations in the component and base assets.

severinbeauvais: > I think they were not persisted in this component, so this code just fixed it.

I'm not sure what you mean by "persisted". This component should use the base classes when it's imported in Storybook or any other UI.

lewischenstudio: This component isn't displaying the error bar and texts.

src/components/certify/Certify.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div id="AR-step-4-container">
+  <div id="AR-step-4-container" :class="{'invalid-section': invalidSection}">
```
lewischenstudio, line 2: The 'invalid-section' is missing from the original code.

severinbeauvais:
> Steps I did on my local environment:
> 
> * I ran storybook to verify the "invalidSection" variable is working properly.
> * I used Node 12 and ran `npm i` successfully without affecting the `package-lock.json` file
> * I ran `npm run storybook:build` which updated too many files inside `doc` folder, which wasn't committed.

Good.
Good.
Not good. Usually it's just a handful of files. Did you build SB under Node 12?

If you can't sort out the SB files, I'd prefer that you didn't commit any of them, instead of checking in a mess for someone else to fix later. (Someone else can try rebuilding SB and checking in the handful of changes. In the meantime, it means people can't use SB to see your changes. But I think this is the lesser of 2 evils.)

src/components/certify/Certify.vue
severinbeauvais: The errors classes should be global, in this project (for Storybook to display your component) and in any UI project that uses this shared component.

I do not want to see local declarations of the error classes.

src/components/certify/Certify.vue
severinbeauvais: `invalid-section` probably should not be in the local code. It is an error class and, as such, should be declared globally (in assets folder) in every UI where this component is used (Storybook, Filings UI, etc). Probably it's already declared in most UIs.

severinbeauvais:
@cameron-freshworks I added you as a reviewer since I'm away most of next week, and you know this project well. Thanks :)

src/components/certify/Certify.vue
cameron-freshworks: Agree with Severin, typically this is applied to a wrapper of sorts wherever locally implemented or if no section wrapper then on the component itself. 
ie `<article :class="{'invalid-section': invalidSection}"><component /></section>`
ie `<Certify :class="{'invalid-section': invalidSection}" />`

src/components/certify/Certify.vue
cameron-freshworks: Definitely agree. 
This class will be overridden by the local definition either way, so it makes sense to add this class to the storybook assets styles (if it doesn't already exist), since this class will only be used for the storybook instance afaik.

lewischenstudio: @cameron-freshworks The classes "invalid-section" and "error-text" are available in the base.scss file: [here](https://github.com/bcgov/bcrs-shared-components/blob/5c1623427f435c7b730d04c79aa06e82344a24ff/src/assets/styles/base.scss#L286). In the original implementation, the style "error-text" didn't take place (turn the text into red) when the variable "invalidSection" is false.

cameron-freshworks: 🤔 
To that point, isn't that expected. Shouldn't the class binding only take effect when the variable is `true`?

lewischenstudio: I am guessing the reason is because of"error-text" is added to to "title-label" for the "Legal Name" and "certify-stmt" for the comments "I, [Legal Name] ...". The local class becomes ".title-label.error-text" and overrides "error-text" from the globally defined style.

lewischenstudio: This is the reason why I created this ticket on Friday and wanted to give it a quick fix. Create UI is using this component, and the error bar and texts are not turning red when it sends "invalidSection = true" to this component.

src/components/certify/Certify.vue
lewischenstudio: Below are two places how the Create UI uses this component: [here](https://github.com/bcgov/business-create-ui/blob/c0ab60baa549c77d87d5a29482c8d6a7f6b120b7/src/views/Registration/RegistrationReviewConfirm.vue#L83-L94) and [here](https://github.com/bcgov/business-create-ui/blob/main/src/components/common/Certify.vue). I understand we can do `<Certify :class="{'invalid-section': invalidSection}" />`, but this style error bug still exists. 

image

image

Because the source code for this component is in BCRS, I created this ticket to fix this bug in this project.

src/components/certify/Certify.vue
cameron-freshworks: Interesting. 

Would this be a scenario where the `color` could be defined with `!important` locally. 
Or potentially this `error-text` defined in the `overrides`?

lewischenstudio: This second image shows the result using the original code as below (I removed the coded local css style):

image

cameron-freshworks:
@lewischenstudio Sorry Lewis, since i just jumped into this, could i get a little context on the bug and where this occurs? 
From a quick look into some of the local instances of this component, i can see that border and label colour take effect when prompted, so i'm confused as to what the bug is?

From Dev: 

Screen Shot 2022-08-22 at 10 32 43 AM

lewischenstudio:
> 

Oh wow! The error bar and texts DO take place when the checkbox is unchecked on Dev. I checked it on Friday, and they didn't take place. That's why I created this ticket. Maybe you are right. Maybe I didn't check it on Dev but on my local work instead, which didn't have the correct implementation to trigger the error bars. It's likely that way. I will close this PR and ticket and look into my implementation. Thanks a lot @cameron-freshworks @severinbeauvais

severinbeauvais:
PS Tickets need to be approved before you work on them.

cameron-freshworks:
> > 
> 
> Oh wow! The error bar and texts DO take place when the checkbox is unchecked on Dev. I checked it on Friday, and they didn't take place. That's why I created this ticket. Maybe you are right. Maybe I didn't check it on Dev but on my local work instead, which didn't have the correct implementation to trigger the error bars. It's likely that way. I will close this PR and ticket and look into my implementation. Thanks a lot @cameron-freshworks @severinbeauvais

No problem at all sir, thanks for looking into this!
In oder to get this invalid state, i tried to submit the filing in an invalid state which was prompting the validations on the review step 👍

lewischenstudio:
> PS Tickets need to be approved before you work on them.
> 
> My fault I didn't catch this earlier.

You are right. I will be very careful next time in creating a ticket. I thought this is a quick fix and I checked it on Dev, but turned out I wasn't looking at the right environment. Sorry for the trouble.

lewischenstudio:
> > > 
> > 
> > 
> > Oh wow! The error bar and texts DO take place when the checkbox is unchecked on Dev. I checked it on Friday, and they didn't take place. That's why I created this ticket. Maybe you are right. Maybe I didn't check it on Dev but on my local work instead, which didn't have the correct implementation to trigger the error bars. It's likely that way. I will close this PR and ticket and look into my implementation. Thanks a lot @cameron-freshworks @severinbeauvais
> 
> No problem at all sir, thanks for looking into this! In oder to get this invalid state, i tried to submit the filing in an invalid state which was prompting the validations on the review step 👍

That is right. Thanks again for verifying it on Dev. ❤️

PR #145, opened 2022-08-24
13021 - Changes for Special Resolution

seeker25:
*Issue #:* /bcgov/entity#13021
*Description of changes:*

Move over association-types.ts from business-edit-ui.
Enhance alteration-interface.ts.
Change import for business-lookup-interfaces.ts to bcrs-shared-components, this works nicer when using npx link.
Add in change-of-name-interface.ts, person-interface.ts, special-resolution-interface.ts.


src/interfaces/business-lookup-interfaces.ts
```diff
@@ -1,4 +1,4 @@
-import { CorpTypeCd, EntityStates } from '@/enums'
+import { CorpTypeCd, EntityStates } from '@bcrs-shared-components/enums'
```
seeker25, line 1: `npx link` had issues with @/enum it wouldn't resolve easily, this syntax was used in another file.

src/interfaces/special-resolution-interface.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { PersonIF } from './person-interface'
+
+export interface SpecialResolutionIF {
+  meetingDate?: string
+  resolution: string
+  resolutionDate?: string
+  signingDate?: string
+  signatory?: PersonIF
+  resolutionConfirmed?: boolean
```
seeker25, line 14: Included here: https://github.com/bcgov/business-schemas/pull/107

src/interfaces/change-of-name-interface.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * A filing's business object from the API. See:
+ * https://github.com/bcgov/business-schemas/blob/master/src/registry_schemas/schemas/change_of_name.json
+ * This requires either nameRequest or legalName:
+*   "anyOf": [
```
seeker25, line 5: Not sure if this should be oneOf instead?

seeker25:
@severinbeauvais @lewischenstudio @lambert-alex I don't have permissions on this repo, but could you review this?

src/enums/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bcrs-shared-components/enums",
-  "version": "1.0.19",
+  "version": "1.0.20",
```
severinbeauvais, line 3: Don't change this. The Lerna command in step 7 does this for you.

Please ensure you follow exactly the steps in the [README](https://github.com/bcgov/bcrs-shared-components) otherwise the versions can get messed, the branches can get messed up, or StoryBook can get messed up.

src/interfaces/court-order-interface.ts
```diff
@@ -2,8 +2,9 @@
  * See:
  * https://github.com/bcgov/business-schemas/blob/master/src/registry_schemas/schemas/court_order.json
  */
-export interface CourtOrderIF{
+export interface CourtOrderIF {
   fileNumber: string
   orderDate: string
-  effectOfOrder: string
+  effectOfOrder: string,
```
severinbeauvais, line 8: If you're in here again, remove the trailing comment. It's not needed and generally our code style is to leave it out so it looks cleaner.

src/interfaces/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bcrs-shared-components/interfaces",
-  "version": "1.0.39",
+  "version": "1.0.40",
```
severinbeauvais, line 3: Revert this change.

src/interfaces/person-interface.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface PersonIF {
+  givenName?: string
+  familyName?: string
+  additionalName?: string
+  middleInitial?: string
+  email?: string
+  [propName: string]: any // excess properties
+}
+
+export const EmptyPersonIF: PersonIF = {
```
severinbeauvais, line 15: This is an object, not an IF. Please rename it.

src/interfaces/person-interface.ts
```diff
@@ -0,0 +1,19 @@
...
+  givenName: '',
+  familyName: '',
+  additionalName: null
```
severinbeauvais, line 18: This is probably OK but watch out for the following:
- you're defining some optional properties
- you're inconsistently defining some string properties as empty strings, and other as null
- the remaining properties are undefined, meaning that changes to them will not trigger reactivity 

What to define in this Empty Person object depends on how it's used, which is subject not objective, which is not ideal. I'd like to define a best practise for such empty objects. Any thoughts?

Obviously, required properties should be defined. There aren't any here.

Should all optional properties be defined?

Should primitive types (eg, string, number) be assigned empty values or null?

src/interfaces/special-resolution-interface.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface SpecialResolutionIF {
+  meetingDate?: string
+  resolution: string
+  resolutionDate?: string
+  signingDate?: string
+  signatory?: PersonIF
+  // Not in API schema, but used in UI
+  resolutionConfirmed?: boolean
+  [propName: string]: any // excess properties
```
severinbeauvais, line 16: Again, I prefer to get rid of these and fix the code/IF instead.

src/interfaces/person-interface.ts
seeker25: I think I'll leave this out of `bcrs-shared-components`

severinbeauvais (approved):
Looks OK.

Some of the code things I commented on may have been carried over from existing code, but please fix anyway.

As per the readme, you also need to rebuild Storybook and check it its changes.

src/interfaces/court-order-interface.ts
seeker25: Whoops, good eye

src/interfaces/person-interface.ts
severinbeauvais: Object has been removed. This conversation is now obsolete.

src/interfaces/person-interface.ts
severinbeauvais: Object has been removed. This conversation is now obsolete.

severinbeauvais:
@seeker25 Please rebuild and commit the changed Storybook files as per my comment above.

Note that only project admins can complete steps 14 and 15 -- I can do it once SB is updated.

PR #149, opened 2022-09-08
12751 + 13510: added label props to NatureOfBusiness

severinbeauvais:
*Issue #:* bcgov/entity#12751 and bcgov/entity#13510

*Description of changes:*
- added editLabel prop
- added editedLabel prop
- added new props to stories
- updated StoryBook


src/components/nature-of-business/NatureOfBusiness.vue
```diff
@@ -113,7 +113,6 @@
                 <v-icon small color="primary">mdi-pencil</v-icon>Change</v-btn>
```
severinbeauvais, line 113: Note that dropdown button remains as "Change" (since we're changing an edit)... confirmed with Yui.

lewischenstudio (approved):
It looks great!

PR #153, opened 2022-10-17
13875 NOB emits valid upon loading a save

lambert-alex:
*Issue #:* /bcgov/entity#13875

*Description of changes:*
- Upon resuming a save NOB emits valid on saved changes

### DIFF
--- a/src/components/nature-of-business/NatureOfBusiness.vue
+++ b/src/components/nature-of-business/NatureOfBusiness.vue
@@ -242,6 +242,7 @@ export default class NatureOfBusiness extends Vue {
   @Watch('haveNaics', { immediate: true })
   private onHaveNaicsChanged (val: boolean): void {
     this.state = val ? States.SUMMARY : States.INITIAL
+    this.emitValid(this.haveNaics)
   }
 
   /** Called when this form's validity has changed. */


lambert-alex (commented):
With this change I should just be able to update the bcrs package in the edit ui and the component will update validation upon reopening a save

src/components/nature-of-business/NatureOfBusiness.vue
```diff
@@ -242,6 +242,7 @@ export default class NatureOfBusiness extends Vue {
   @Watch('haveNaics', { immediate: true })
   private onHaveNaicsChanged (val: boolean): void {
     this.state = val ? States.SUMMARY : States.INITIAL
+    this.emitValid(this.haveNaics)
```
severinbeauvais, line 245: I'm hesistant about this solution since validity also depends on whether this component is expanded.

Would making line 263 emit an initial value work instead?

lewischenstudio:
I checked this ticket on Edit Ui, but it's not showing an error after saving the draft. I wonder if this ticket is for Create UI.

severinbeauvais:
Good catch. Yes, the ticket refers to a registration (Create UI).

src/components/nature-of-business/NatureOfBusiness.vue
lambert-alex: It does not unfortunately, immediate checks before the filing gets build. I think that this should be safe however since once you chose a NAICS value then you can't set it back to an empty value, so it should be safe.

lambert-alex:
> I checked this ticket on Edit Ui, but it's not showing an error after saving the draft. I wonder if this ticket is for Create UI.

Yes it does for registration filings, I never changed the ticket, I'll do that thanks for the reminder!

src/components/nature-of-business/NatureOfBusiness.vue
severinbeauvais: Does this component work correctly in Edit UI? What's different between Create UI and Edit UI?

lambert-alex: It does however the way the components is used differently. In the Edit UI each component sets validity where as in the Create UI the Registration view sets the validity.

severinbeauvais: I mean, some sort of change in the local, intermediate component NatureOfBusiness.vue?

lambert-alex: Doing it that way would make NOB the only component to validate like that in registration. Which might not be great

Another solution I could do is use the updated function in the NOB component so it emits the NAICS code after the component has been loaded?

ie ```updated () { this.emitValid(!!this.getRegistration.naics) }``` in the create UI

severinbeauvais: Does the shared NOB component emit an initial status? Is it incorrect? If so, why?

What's the actual problem here -- is the initial validation status incorrect? Not being updated?

lambert-alex: Sorry for not getting back to you.

So I have been going about it the wrong way. The issue is that the emit https://github.com/bcgov/business-create-ui/blob/7b256588cb28eeb800d7ea807bac02a740a0ddbd/src/views/Registration/RegistrationDefineBusiness.vue#L34
is not running this function https://github.com/bcgov/business-create-ui/blob/7b256588cb28eeb800d7ea807bac02a740a0ddbd/src/views/Registration/RegistrationDefineBusiness.vue#L204

I am trying to figure out why

lambert-alex: So what is happening is the NOB components only emits once if there is a change to the component and the validity is local to the view and then validity is evaluated upon building the filing. The edit ui gets around this issue by having the validity set and stored in state so when the filing is built the validity of the NOB is loaded as well.

The two ways that I have found to fix this are around this are. 1. Trigger the emit like above. 2. Have a function like mounted or updated trigger the emit when NOB is rendered. This is what StartDate does

severinbeauvais: I had a quick look at this locally. The problem I saw is that the shared NOB component does not emit an initial "valid" event, so RegistrationDefineBusiness.vue doesn't get an initial value (the value stays False).

This seems to work correctly with this change:
image

What do you think?

severinbeauvais (approved):

PR #158, opened 2023-01-20
14803 add approval type component

jonathan-longe:
*Issue #:* /bcgov/entity14803

#### Default
image

#### Default with `isCourtOrderOnly` = true
image

#### Default with `isConversionToFullRestoration` = true
image

#### Draft with `approvedByRegistrar` = true
image

#### Draft with `draftCourtOrderNumber` = '99-1234567'
image

#### With Validation error shown
image


src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,169 @@
...
+          />
+          <v-form ref="courtNumRef" id="court-num-form" v-model="valid" class="mt-4 ml-8">
+            <v-text-field
+              @input="courtOrderNumberChanged"
+              id="court-order-number-input"
+              v-model="courtOrderNumber"
+              label="Court Order Number"
+              :rules="courtOrderNumRules"
+              :disabled="approvalTypeSelected === ApprovalTypes.VIA_REGISTRAR"
+              @update:error="emitValidationError"
```
severinbeauvais, line 22: I prefer to use the full syntax for the event handler, so it's clear that it's a method and not a property, ie, `emitValidationError()`.

The downside is that you have to explicitly pass the event data (`$event`), which is inherent in the first synatax.

Not a deal-breaker.

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  /** Draft court order number. */
+  @Prop({ default: '' }) readonly draftCourtOrderNumber!: string
+
+  /** Whether approved by the registrar. */
+  @Prop({ default: false }) readonly approvedByRegistrar!: boolean
+
+  /** text used in radio options . */
+  @Prop({ default: false }) readonly isConversionToFullRestoration!: boolean
```
severinbeauvais, line 64: I counsel against identifying a specific filing type here... it would be better to use a flag named after the desired behaviour. (This component should not know what's going on in its parent.)

Also, comment is outdated.

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,169 @@
...
+export default class ApprovalType extends Vue {
+  // Refs
+  $refs!: Vue['$refs'] & {
+    courtNumRef: FormIF
+  }
+
+  ApprovalTypes = {
+    VIA_COURT_ORDER: 'VIA COURT ORDER',
+    VIA_REGISTRAR: 'VIA REGISTRAR'
+  }
```
severinbeauvais, line 55: I believe this is an object, not an enum. Seems to work though?

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,169 @@
...
+      (v: string) => (!v || !(v.length < 5)) || 'Court order number is invalid',
+      (v: string) => (!v || !(v.length > 20)) || 'Court order number is invalid',
+      (v: string) => !!v || 'A Court Order number is required'
+    ]
+  }
+
+  private getRadioText (option: string): string {
+    let restorationWording = 'restoration'
+    if (this.isConversionToFullRestoration) {
+      restorationWording = 'conversion to full restoration'
```
severinbeauvais, line 133: If it's just a label changing, IMO you should pass the label string as a prop. It will keep this component more generic.

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,169 @@
...
+    }
+    return '[error]'
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+#court-order-label {
```
severinbeauvais, line 148: Unused CSS?

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,169 @@
...
+
+  private setCourtNumberRules (): void {
+    // Apply TextField rules
+    this.courtOrderNumRules = [
+      (v: string) => (!v || !/^\s/g.test(v)) || 'Invalid spaces', // leading spaces
+      (v: string) => (!v || !/\s$/g.test(v)) || 'Invalid spaces', // trailing spaces
+      (v: string) => (!v || !(v.length < 5)) || 'Court order number is invalid',
+      (v: string) => (!v || !(v.length > 20)) || 'Court order number is invalid',
+      (v: string) => !!v || 'A Court Order number is required'
+    ]
```
severinbeauvais, line 127: This is fine, but FYI, another way this can be done is in the component itself, like this:
```
   :rules='enableValidation ? courtNumberRules : []
```

Where enableValidation is a flag you togle and courtNumberRules is a readonly const.

severinbeauvais: Do these rules ever need to be cleared?

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,169 @@
...
+    </v-row>
+  </div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { Component, Emit, Prop, Watch } from 'vue-property-decorator'
+import { FormIF } from '@bcrs-shared-components/interfaces'
+import { types } from "sass";
+import Null = types.Null;
```
severinbeauvais, line 43: What are these 2 imports?

src/components/approval-type/package.json
```diff
@@ -0,0 +1,11 @@
...
+  "name": "@bcrs-shared-components/approval-type",
+  "version": "0.0.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@bcrs-shared-components/interfaces": "^1.0.47",
+    "vue-property-decorator": "^9.1.2"
+  }
+}
```
severinbeauvais, line 11: 👍

tests/unit/ApprovalType.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      approvedByRegistrar,
+      isConversionToFullRestoration
+    },
+    vuetify,
+    localVue
+  })
+}
+
+describe('Initialize ApprovalType component', () => {
+  it('loads the component', async () => {
```
severinbeauvais, line 33: Not async.

Not a deal-breaker.

tests/unit/ApprovalType.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      .toEqual('This conversion to full restoration is approved by court order.')
+    wrapper.destroy()
+  })
+
+  it('component emits events when court order number entered', async () => {
+    const wrapper: Wrapper<ApprovalType> = createDefaultComponent()
+    // Input text into text-field
+    const input = wrapper.find('#court-order-number-input')
+    input.setValue('89123456')
+    await Vue.nextTick()
```
severinbeauvais, line 67: I'm not certain if it will work here, but sometimes you can simply await the setValue() above.

Sometimes you need to wait for the next timeslice (remember, JS is single-threaded) so that validation can take place and set the properties you expect to see changed.

severinbeauvais (commented):
Looks really good. Clearly you got into the depths of everything :)

src/components/approval-type/ApprovalType.vue
jonathan-longe: Ah, good thought, I'll change it.

src/components/approval-type/ApprovalType.vue
jonathan-longe: Good call, will fix

src/components/approval-type/ApprovalType.vue
jonathan-longe: Good call, I'll fix

tests/unit/ApprovalType.spec.ts
jonathan-longe: Good to know.  If it's okay for you, I'll leave this as is.

tests/unit/ApprovalType.spec.ts
jonathan-longe: Good catch.  I'll fix it

src/components/approval-type/ApprovalType.vue
jonathan-longe: Nice catch, removed

src/components/approval-type/ApprovalType.vue
jonathan-longe: I've removed what I think are the unnecessary import statements.  Let me know if you see any that are still not required.

src/components/approval-type/ApprovalType.vue
jonathan-longe: Good to know.  I'll leave this as is for now, but I appreciate the feedback.

src/components/approval-type/ApprovalType.vue
severinbeauvais: Sometimes imports are automatically added by an IDE extension when you (accidentally) type in some things. I bet that's what these are.

src/components/approval-type/ApprovalType.vue
severinbeauvais: Do these rules ever need to be cleared?

src/components/approval-type/ApprovalType.vue
severinbeauvais: Thanks for update.

Same thing on line 8 actually.

src/components/approval-type/ApprovalType.vue
severinbeauvais: I bet this would work as an enum (`enum ApprovalTypes = { etc`).

src/components/approval-type/ApprovalType.vue
severinbeauvais: Looks good.

severinbeauvais (approved):
Please wait for another review before merging. I should not be the primary or only reviewer.

jonathan-longe:
> Please wait for another review before merging. I should not be the primary or only reviewer.

Thanks @severinbeauvais !

src/components/approval-type/ApprovalType.stories.ts
```diff
@@ -0,0 +1,37 @@
+import { ApprovalType } from './index'
+
+export default {
+  title: 'component/ApprovalType',
+  component: ApprovalType,
+  argTypes: {
+  }
```
cameron-eyds, line 7: Super nit picky, but probably doesn't need a new line. ie `argtypes: {}`

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,157 @@
...
+            <v-text-field
+              @input="courtOrderNumberChanged"
+              id="court-order-number-input"
+              v-model="courtOrderNumber"
+              label="Court Order Number"
+              :rules="courtOrderNumRules"
+              :disabled="approvalTypeSelected === ApprovalTypes.VIA_REGISTRAR"
+              @update:error="emitValidationError($event)"
+              filled
+            />
```
cameron-eyds, line 24: Once again, super nitpicky, we try to lead with the id, followed by properties/bound properties and subsequently event handlers. 
So in this case, maybe just move the `input` event to bottom with `update` etc. 
Not required at all, just general practice.

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,157 @@
...
+          </v-form>
+          <!-- REGISTRAR radio button -->
+          <v-radio id="registrar-radio" class="mb-0 pt-2"
+             :label="getRadioText(ApprovalTypes.VIA_REGISTRAR)"
+             :value="ApprovalTypes.VIA_REGISTRAR"
+             v-if="!isCourtOrderOnly"
```
cameron-eyds, line 30: similar  to the above comment, leading with a conditional (when present) is what we've done in the past, with regards to arrangement of the component props. 
ie `<v-radio v-if="...`

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,157 @@
...
+    this.$emit('emitValid', status)
+    return status
+  }
+
+  private emitValidationError (event): void {
+    this.$emit('emitValid', !event)
+  }
+
+  @Emit('emitRadioButtonChange')
+  private radioButtonChanged (event): void {
```
cameron-eyds, line 98: types on the arguments might be helpful here.

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,157 @@
...
+      (v: string) => (!v || !/\s$/g.test(v)) || 'Invalid spaces', // trailing spaces
+      (v: string) => (!v || !(v.length < 5)) || 'Court order number is invalid',
+      (v: string) => (!v || !(v.length > 20)) || 'Court order number is invalid',
+      (v: string) => !!v || 'A Court Order number is required'
+    ]
+  }
+
+  private getRadioText (option: string): string {
+    if (option === this.ApprovalTypes.VIA_COURT_ORDER) {
+      return 'This ' + this.filingType + ' is approved by court order.'
```
cameron-eyds, line 130: No action required but an alternative to the string concatenation is use template strings. 
ie `This ${this.filingType} is approved by court order.` 
Just wrap in back ticks.

src/components/approval-type/ApprovalType.vue
severinbeauvais: If it were me, I'd name the first one "valid" and the second one "value" (or val).

cameron-eyds (approved):
Looks good to me! 
Just some minor comments 👍

JazzarKarim (approved):
Looks good to me!

ketaki-deodhar (approved):
Lots of learning here for me :)

src/components/approval-type/ApprovalType.stories.ts
jonathan-longe: Nice, I'll fix it!

src/components/approval-type/ApprovalType.vue
jonathan-longe: Thanks for letting me know.  I'll make the changes you suggest.

src/components/approval-type/ApprovalType.vue
jonathan-longe: Makes sense. I'll fix it.

src/components/approval-type/ApprovalType.vue
jonathan-longe: I agree, that's better.

src/components/approval-type/ApprovalType.vue
jonathan-longe: Nice, that's prettier for sure.  I'll fix it.

src/components/approval-type/ApprovalType.vue
jonathan-longe: prefixing ApprovalTypes with `enum` causes the unit tests to fail.  Are you okay if I leave this as is?

src/components/approval-type/ApprovalType.vue
jonathan-longe: No, not as far as I know.  Do I need a public clear() method?  Is that what you're suggesting?

src/components/approval-type/ApprovalType.vue
severinbeauvais: Yup.

src/components/approval-type/ApprovalType.vue
severinbeauvais: I was thinking that you might need to set the rules to [] to disable them.

But maybe this is "once enabled, always enabled".

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,157 @@
...
+
+  /** Triggers the form validation. */
+  public validate (): boolean {
+    let status = this.$refs.courtNumRef.validate()
+    this.$emit('emitValid', status)
+    return status
+  }
+
+  private emitValidationError (event: boolean): void {
+    this.$emit('emitValid', !event)
```
severinbeauvais, line 94: We usually use the `@Emit` decorator instead of this, but it's valid.

src/components/approval-type/ApprovalType.vue
```diff
@@ -0,0 +1,157 @@
...
+    }
+    if (event === this.ApprovalTypes.VIA_COURT_ORDER) {
+      this.setCourtNumberRules()
+    }
+    this.validate()
+  }
+
+  @Emit('emitCourtNumberChange')
+  private courtOrderNumberChanged (event): void {
+    Vue.set(this, 'approvalTypeSelected', this.ApprovalTypes.VIA_COURT_ORDER)
```
severinbeauvais, line 111: Why can't this be a simple assignment?

jonathan-longe: My understanding is that simple assignments are not reactive.

src/components/approval-type/ApprovalType.vue
jonathan-longe: Yes, but the `@Emit` decorator doesn't allow me to change the value of the event.  On line 22, I call `@update:error="emitValidationError($event)"`  If there's a better way to do this, please let me know.

src/components/approval-type/ApprovalType.vue
severinbeauvais: This PR is water under the bridge, but I'm fairly certain simple assignments _are_ reactive;  we use them all the time.

My understanding is that Vue.set adds a property to an existing reactive object (ie, related to deep reactivity). I think it's _probably_ not needed here.

src/components/approval-type/ApprovalType.vue
severinbeauvais: If you return the value in an emit method, it should pass the value along, eg.

```
@Emit('myEvent')
protected emitMyEvent (myEvent: string): string {
   return myEvent
}
```

Or even simpler syntax:
```
@Emit('myEvent') protected emitMyEvent (myEvent: string): void {}
```

PR #160, opened 2023-02-01
14797 relationships panel

JazzarKarim:
*Issue #:* bcgov/entity#14797

*Description of changes:*
- Created Relationships Panel

Default:
default

If no relationship is selected:
none selected

- The component has one prop and two emitters:
   - Prop: Draft relationships array
   - 1st emitter: Selected relationships (Array)
   - 2nd emitter: Check and validate if at least one relationship is selected (Boolean)


src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
+<template>
+  <v-card flat id="relationship-style">
+    <div class="form__row three-column mt-3">
```
severinbeauvais, line 3: Please verify that your component looks correct in Storybook. For example, I believe "three-column" is not declared as a base style in this repo -- you can add it.

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
+<template>
+  <v-card flat id="relationship-style">
```
severinbeauvais, line 2: The id of this component should match its name, please.

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
...
+    <div class="form__row three-column mt-3">
+      <v-card flat rounded="sm" class="relationship_title gray-card px-4">
```
severinbeauvais, line 4: Please rename this class. It's not a title. Also we don't use snake-case for classnames. You could name is "relationship-vcard".

severinbeauvais: You'll need to add "gray-card" class to the styles (assets) in this repo.

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
...
+  private relationshipRules = []
+
+  // Relationship Types Enum for the template
+  RelationshipTypes = {
+    COURT_ORDERED_PARTY: 'Court Ordered Party',
+    DIRECTOR: 'Director',
+    HEIR_LEGAL_REP: 'Heir or Legal Representative',
+    OFFICER: 'Officer',
+    SHAREHOLDER: 'Shareholder'
+  }
```
severinbeauvais, line 60: You should create a new, shared enum for this (and specify it as a dependency for THIS shared component).

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
...
+
+  /**
+   * Called when component is mounted.
+   * Automatically check all previously checked relationships (if any) when user continues a draft.
+   */
+  mounted (): void {
+    if (this.draftRelationships !== null) {
```
severinbeauvais, line 67: This can never be null, because the default prop valud is used if the parent passes down a null or undefined value.

But it could be an empty array.

Do we require a default of "heir legal rep"? Maybe the prop default could be `[]` instead?

src/components/relationships-panel/RelationshipsPanel.vue
JazzarKarim: Component looks correct in Storybook. I will move three-column to the base style.

src/components/relationships-panel/RelationshipsPanel.vue
severinbeauvais: In the story, you could test passing an array with one or more values.

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
...
+</template>
+<script lang="ts">
+
+import Vue from 'vue'
+import { Component, Watch, Emit, Prop } from 'vue-property-decorator'
+
+@Component({})
+export default class RelationshipsPanelShared extends Vue {
+  /** Draft restoration relationships */
+  @Prop({ default: ['Heir or Legal Representative'] }) readonly draftRelationships!: []
```
severinbeauvais, line 47: This _may_ work but it's not the correct way to set a default array (or object). Should be:
```
@Prop({ default: () => [] }) readonly draftRelationships!: RelationshipTypes[]
```

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
...
+  private relationshipsValid (event: boolean): boolean {
+    return event
+  }
+
+   /**
+    * Emit events whenever the relationships array is changed (validation and the selected relationships array).
+    * Change the tooltip text to red (error) if no relationships are selected.
+    */
+   @Watch('selectedRelationships')
+  private setRelationships (val) {
```
severinbeauvais, line 98: Indent issue. Make sure you run `npm run lint` locally before committing code.

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
...
+import { Component, Watch, Emit, Prop } from 'vue-property-decorator'
+
+@Component({})
+export default class RelationshipsPanelShared extends Vue {
+  /** Draft restoration relationships */
+  @Prop({ default: ['Heir or Legal Representative'] }) readonly draftRelationships!: []
+
+  // Local properties
+  private selectedRelationships = []
+  private relationshipRules = []
```
severinbeauvais, line 51: It would be great to have types on these.

ChrisGabel23 (approved):
Subject to the resolution of the other comments.

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
...
+
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+#relationship-style {
+  padding-bottom: 1rem;
+  padding-left: 1rem;
+  margin-right: 1.5rem;
```
severinbeauvais, line 117: For more layout styles, I have found it's better to use the [Vuetify spacing classes](https://vuetifyjs.com/en/styles/spacing/) right in the template -- when fixed or refactoring layouts/whitespace (something we have to do every once in a while), it's _way_ easier to see and tweak the classes in the template instead of having to look for the class and whatever it declares.

src/components/relationships-panel/RelationshipsPanel.vue
severinbeauvais: See my comment below (gray-card).

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
...
+  line-height: 1.2rem;
+}
+
+.gray-card {
+  background-color: rgba(0, 0, 0, 0.06);
+  margin-right: 3rem;
```
severinbeauvais, line 123: `mr-12`

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,141 @@
...
+}
+
+.form__row.three-column {
+  display: flex;
+  flex-flow: row nowrap;
+  align-items: stretch;
+  margin-right: -0.5rem;
+  margin-left: -0.5rem;
```
severinbeauvais, line 131: `ml-n2`

src/components/relationships-panel/package-lock.json
```diff
@@ -0,0 +1,12 @@
...
+  "version": "0.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@bcrs-shared-components/relationships-panel",
+      "version": "0.0.0"
+    }
+  }
+}
```
severinbeauvais, line 12: The other components in this repo don't have a package lock file, so yours shouldn't either.

src/components/relationships-panel/RelationshipsPanel.vue
JazzarKarim: Created a relationship-types enum. Added the dependency.

src/components/relationships-panel/RelationshipsPanel.vue
JazzarKarim: I kept this as is. This is because when using it in the create-ui, if limited restoration draft was continued, the restoration relationships getter (passed as a prop), is going to return null (getRestoration.relationships). Hence, in this case, when user selects full restoration, I will just populate my selected relationships array with heir legal rep. I tested this in create-ui.

src/components/relationships-panel/RelationshipsPanel.vue
JazzarKarim: Fixed. Now it's:
@Prop({ default: () => [RelationshipTypes.HEIR_LEGAL_REP] }) readonly draftRelationships!: RelationshipTypes[]

src/components/relationships-panel/RelationshipsPanel.vue
JazzarKarim: Good catch. Fixed!

src/components/relationships-panel/RelationshipsPanel.vue
JazzarKarim: Added types to both arrays. Fixed!

src/components/relationships-panel/package-lock.json
JazzarKarim: Removed the file.

src/components/relationships-panel/RelationshipsPanel.vue
JazzarKarim: Fixed. I cleaned up my styles and code. I removed the unnecessary div. I also removed the gray-card. Now, the styles are specific to this component via relationships-panel and the relationships-vcard class.

src/components/relationships-panel/RelationshipsPanel.vue
severinbeauvais: I believe that, if getRestoration.relationships returns null, this component will use the default prop value. So this.draftRelationships can never be null.

I think that a default of `[]` is more generic, in case this component is used elsewhere. The parent should set any initial values.

One more thing to look at: getRestoration.relationships should never return null. It should be initialized (in state model) as an empty array. And when resuming a draft and parsing it, you should do either of these:
`if (filing.restoration.relationships) this.setRestorationRelationships(filing.restoration.relationships)`
or
`this.setRestorationRelationships(filing.restoration.relationships || [])`

I prefer the first since it means we set a default value in only 1 place (store init).

What do you think?

severinbeauvais:
> If no relationship is selected: none selected

Do you show error message text if no relationships are selected?

src/components/relationships-panel/RelationshipsPanel.vue
severinbeauvais: For Storybook purposes, you can have one story where you pass in one (or more) relationships to initially display. (And you should test the defaul props -- which you do already.)

JazzarKarim:
> > If no relationship is selected: none selected
> 
> Do you show error message text if no relationships are selected?

I spoke with Ethan about the design. As per the UXPin, no text is going to be shown. Instead, the tooltip text and the relationships will be marked as red (error) as follows:

uxpin

src/components/relationships-panel/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@bcrs-shared-components/relationships-panel",
+  "version": "0.0.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@bcrs-shared-components/enums": "^1.0.26",
```
severinbeauvais, line 8: You need to rebase. This is up to v1.0.27 now.

src/enums/relationship-types.ts
```diff
@@ -0,0 +1,7 @@
+export enum RelationshipTypes {
+    COURT_ORDERED_PARTY = 'Court Ordered Party',
+    DIRECTOR = 'Director',
+    HEIR_LEGAL_REP = 'Heir or Legal Representative',
+    OFFICER = 'Officer',
+    SHAREHOLDER = 'Shareholder',
+  }
```
severinbeauvais, line 7: Looks good!

Please also update the index file in the enums folder, as you will be publishing an updated version of the enums package as part of this PR.

src/components/relationships-panel/package.json
JazzarKarim: Yes, I'm already on it!

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,112 @@
...
+        </v-col>
+      </v-row>
+    </v-card>
+  </v-card>
+</template>
+<script lang="ts">
+
+import Vue from 'vue'
+import { Component, Watch, Emit, Prop } from 'vue-property-decorator'
+import { RelationshipTypes } from '@/enums/relationship-types'
```
severinbeauvais, line 40: You need to import the enum like this (otherwise the project that imports this component will have to provide the enum in its own enum folder):
```
import { RelationshipTypes } from '@bcrs-shared-components/enums'
```

tests/unit/RelationshipsPanel.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    // Officer selected
+    const input = wrapper.find('#officer-checkbox')
+    input.setChecked()
+    await Vue.nextTick()
+
+    expect(wrapper.emitted('emitRelationshipsChanged').pop()[0])
+      .toEqual([RelationshipTypes.HEIR_LEGAL_REP, RelationshipTypes.OFFICER])
+    wrapper.destroy()
+  })
+})
```
severinbeauvais, line 70: Great tests! 👍

tests/unit/RelationshipsPanel.spec.ts
```diff
@@ -0,0 +1,70 @@
+import Vue from 'vue'
+import Vuetify from 'vuetify'
+import { createLocalVue, mount, Wrapper } from '@vue/test-utils'
+import { RelationshipsPanel } from '@/components/relationships-panel'
+import { ApprovalType } from '@/components/approval-type'
+import VueRouter from 'vue-router'
+import { RelationshipTypes } from '@/enums/relationship-types'
```
severinbeauvais, line 7: Don't use enums in unit tests. We want the tests to be fragile and thus to check enum values. That way, an error in the enum may be caught by a test -- otherwise the same error may be cancelled out and not noticed.

src/components/relationships-panel/RelationshipsPanel.vue
severinbeauvais: It's way cleaner. Thanks!

src/components/relationships-panel/RelationshipsPanel.vue
severinbeauvais: You found the right type for the rules -- nice!

src/components/relationships-panel/package.json
severinbeauvais: FYI, when I run the Lerna commands for you, it will auto-update this to 28 since Lerna will also be concurrently publishing the enums package.

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Called when component is mounted.
+   * Automatically check all previously checked relationships (if any) when user continues a draft.
+   */
+  mounted (): void {
+    if (this.draftRelationships.length > 0) {
+      this.selectedRelationships.push(...this.draftRelationships)
+    } else {
+      this.selectedRelationships.push() // This is done to validate component immediately (no need to touch)
```
severinbeauvais, line 61: Is this a fancy way to update the local variable and trigger the watcher? I hadn't thought of that!

(Another way is to call the emitter method here.)

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,112 @@
...
+    return event
+  }
+
+  /**
+   * Emit events whenever the relationships array is changed (validation and the selected relationships array).
+   */
+  @Watch('selectedRelationships')
+  private setRelationships (val) {
+    this.setRelationshipRules()
+    this.$emit('emitRelationshipsChanged', this.selectedRelationships)
```
severinbeauvais, line 90: Just call your emitter method on line 80.

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  // Emit the selected relationships array.
+  @Emit('emitRelationshipsChanged')
+  private relationshipsChanged (): any[] {
+    return this.selectedRelationships
+  }
+
+  // Emit a boolean (validation) which is if at least one relationship is selected.
+  @Emit('emitRelationshipsValid')
```
severinbeauvais, line 79: I think you could simplify the names of the events that are emitted back to the parent. For example, they could just be "changed" and "valid".

severinbeauvais: (Sorry, I wish I had noticed this earlier.)

src/components/relationships-panel/RelationshipsPanel.vue
JazzarKarim: Oh okay I didn't know that. Will do.

tests/unit/RelationshipsPanel.spec.ts
JazzarKarim: Thank you! 🙏

tests/unit/RelationshipsPanel.spec.ts
JazzarKarim: Okay I removed it. Makes a lot of sense.

src/components/relationships-panel/RelationshipsPanel.vue
JazzarKarim: No worries!

src/components/relationships-panel/RelationshipsPanel.vue
severinbeauvais: We do have some components that weren't done this say, so they require an interface/enum/mixin/whatever both here for Storybook and also in the parent project :(  I guess we'll clean them up as we touch them.

severinbeauvais: Er, I don't see any existing shared components that do what I just said we should do. Maybe it doesn't work...?

JazzarKarim: I was checking all the other components that's been already done and I saw many of them depend on shared stuff like an enum or interface like you said!

jonathan-longe (approved):
Good work @JazzarKarim

ketaki-deodhar (approved):

JazzarKarim:
> Good work @JazzarKarim

Thanks Jonathan! 🙏

src/components/relationships-panel/RelationshipsPanel.vue
severinbeauvais: Obsolete comment.

severinbeauvais (approved):
Looks good!

JazzarKarim:
> Looks good!

Glad to hear that. Thanks so much Sev! 😃

PR #161, opened 2023-02-04
14797 Added limited restoration radio panel component

JazzarKarim:

*Description of changes:*
- Created Limited Restoration Months Radio Panel

Default:
Default

Custom Selected:
Custom

Decimal Value Input:
enter valid value

Exceeding Boundaries Input:
boundaries

Empty Input:
empty

- The component has two props and one emitter:
   - 1st Prop: Current Date
   - 2nd Prop: Expiry Date
   - Emitter: The Expiry Date


jonathan-longe (approved):

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,104 @@
...
+   */
+   @Watch('selectMonths')
+  private setCustomMonths (val) {
+    this.customMonths = val === 'customMonths'
+    if (val !== 'customMonths') {
+      this.expiryChanged(this.addMonthsToDate(val, this.currentDate))
+    } else {
+      this.expiryChanged(this.addMonthsToDate(this.numberOfMonths, this.currentDate))
+    }
+  }
```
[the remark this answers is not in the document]
severinbeauvais: I believe Vue attaches watcher to almost everything (that isn't undefined) -- which is how reactivity works. If so then explicit watchers shouldn't hurt anything.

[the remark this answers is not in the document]
jonathan-longe: I'd be very happy to learn more about how Vue works under the hood.  I was basing my comment on the Vue docs that recommend against using watchers.  _"Vue does provide a more generic way to observe and react to data changes on a Vue instance: watch properties. When you have some data that needs to change based on some other data, it is tempting to overuse watch - especially if you are coming from an AngularJS background. However, it is often a better idea to use a computed property rather than an imperative watch callback."_  See:  https://v2.vuejs.org/v2/guide/computed.html#Computed-vs-Watched-Property

In this case, rather than watching selectMonths, I think we should be able to call setCustomMonths when needed.  Not a big deal, just a suggestion for consideration.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,104 @@
...
+  <v-radio-group class="mt-0 pt-0 ml-8" v-model="selectMonths" mandatory>
+    <v-radio id="twentyFour-radio" name="selectMonths" label="Two Years" :value=24 />
+    <v-radio id="eighteen-radio" name="selectMonths" label="18 months" :value=18 />
+    <v-radio id="twelve-radio" name="selectMonths" label="12 months" :value=12 />
+    <v-radio id="six-radio" name="selectMonths" label="6 months" :value=6 />
+    <v-row class="ml-0">
+      <v-radio id="custom-months" name="selectMonths" label="Month(s): " value="customMonths">
+      </v-radio>
+      <v-text-field class="shrink mx-4" type="number" min=1 max=24 :rules="monthRules" v-model="numberOfMonths"
+        :disabled="!customMonths" @change="onMonthsChanged" filled />
```
severinbeauvais, line 11: Try adding prop `dense` to make this component a little shorter.

Also try `hide-details="auto"` to see if that helps.

Also, we've never shown the spin buttons before. You can use `hide-spin-buttons`.

Ref: https://vuetifyjs.com/en/api/v-text-field/

Also, you can try using flexbox to vertically align (center) the radio button and the text field: https://vuetifyjs.com/en/styles/flex/#flex-align

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,104 @@
...
+    </v-row>
+  </v-radio-group>
+</template>
+<script lang="ts">
```
severinbeauvais, line 15: Typically we have a blank line between the template and script (and script and style), and no blank line at the top of the script block.

src/mixins/date-mixin.ts
```diff
@@ -167,4 +167,36 @@ export default class DateMixin extends Vue {
...
+    let currDate = this.yyyyMmDdToDate(dateFrom)
+    let monthDiff = expiryDate.getMonth() - currDate.getMonth()
+    let yearDiff = (12 * (expiryDate.getFullYear() - currDate.getFullYear()))
+    let difference = monthDiff + yearDiff
+    // For example if Jan 31st and March 1, difference is 1 not 2
+    if (expiryDate.getDate() < currDate.getDate()) {
+      difference--
+    }
+    return difference
+  }
```
severinbeauvais, line 201: Make sure you add these to the Date Mixin in Create UI. So, yes, your shared component using THIS mixin when running in Storybook, and Create UI's mixin when running there.

src/mixins/date-mixin.ts
```diff
@@ -167,4 +167,36 @@ export default class DateMixin extends Vue {
...
+  /**
+   * Decrease one date from another and return number of months as the difference.
+   * Dates must be in the "YYYY-MM-DD" format
+   * @example (2023-02-03, 2024-08-03) -> 18
+   * @example (2023-02-03, 2023-04-03) -> 2
+   */
+  subtractDates (dateFrom: string, dateTo: string): number {
+    let expiryDate = this.yyyyMmDdToDate(dateTo)
+    let currDate = this.yyyyMmDdToDate(dateFrom)
+    let monthDiff = expiryDate.getMonth() - currDate.getMonth()
```
severinbeauvais, line 193: If dateFrom or dateTo are undefine/null/empty strings, then expiryDate or currDate will be null, and this line will fail.

I suggest you verify that neither are null, and if they are then return a safe value (or throw an exception, which you should then catch in the caller).

Ditto for line 179.

src/mixins/date-mixin.ts
```diff
@@ -167,4 +167,36 @@ export default class DateMixin extends Vue {
...
+
+  /**
+   * Add a number of months to a date and return "YYYY-MM-DD".
+   * Date must be in the format of "YYYY-MM-DD" and months is a number
+   * @example (3, 2023-02-03) -> "2023-05-03"
+   * @example (18, 2023-02-03) -> "2024-08-03"
+   */
+  addMonthsToDate (months: number, date: string): string {
+    let temp = this.yyyyMmDdToDate(date)
+    temp.setMonth(temp.getMonth() + Number(months))
```
severinbeauvais, line 179: You shouldn't need to type `months` -- it's already a number.

tests/unit/LimitedRestorationPanel.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+describe('Initialize RelationshipsPanel component', () => {
+  it('loads the component', () => {
+    const wrapper: Wrapper<LimitedRestorationPanel> = createDefaultComponent()
+    expect(wrapper.findComponent(LimitedRestorationPanel).exists()).toBe(true)
+    wrapper.destroy()
+  })
+
+  it('No expiry field exists (going from full to limited restoration), Months: 24', async () => {
```
severinbeauvais, line 37: We try (sometimes not very hard) to make this description a sentence, eg, "It is initialized to 24 months if the expiry field is not provided)".

src/mixins/date-mixin.ts
```diff
@@ -167,4 +167,36 @@ export default class DateMixin extends Vue {
 
     return `${dateStr} at ${timeStr} Pacific time`
   }
+
```
severinbeauvais, line 170: Please add unit tests for these new methods.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Set restoration expiry date by adding the custom number of months (1-24) input by user to current date.
+   * Prevent the user from selecting values above 24, lower than 1, and decimals.
+   */
+  onMonthsChanged (): void {
+    if (this.numberOfMonths > 24) {
+      this.numberOfMonths = 24
+    } else if (this.numberOfMonths < 1) {
+      this.numberOfMonths = 1
+    } else if (this.numberOfMonths % 1 !== 0) {
```
severinbeauvais, line 77: I've never seen this before, but I get it. Cool.

However, the JS Number object has a built-in method, isInteger(), that would be easier to understand :)

severinbeauvais: Or just always use `floor` -- won't hurt.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,104 @@
...
+  private monthRules: ((data: string) => void)[] = []
+
+  /**
+   * Called when the component is mounted.
+   * Auto select the previous months input of user (if any).
+   * If there is no previous input, auto select 24 months.
+   */
+  mounted (): void {
+    this.setMonthRules()
+    if (this.expiryDate !== null && this.expiryDate !== undefined) { // Unit tests fail without the first condition.
```
severinbeauvais, line 39: This chec kshould not be needed. You have provided default values for both of these, so they should never be null or undefined here.

However, your default values are not computable, and it's possible the caller passed in non-computable values, so you might want to check that.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Set if custom number of months is selected.
+   * Set expiry date by adding the value of the selected radio buttons to current date.
+   * @param val The value of the selected radio button
+   */
+   @Watch('selectMonths')
+  private setCustomMonths (val) {
+    this.customMonths = val === 'customMonths'
+    if (val !== 'customMonths') {
```
severinbeauvais, line 91: Instead of doing the comparison twice (lines 90 and 91), you could check here whether `customMonths` is False

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,104 @@
...
+    this.monthRules = [
+      (v:string) => !!v || 'Please enter the number of months',
+      (v:string) => (v > 0 && v <= 24) || 'This field must be between 1 to 24', // Keep month between 1 and 24
+      (v:string) => (v % 1 === 0) || 'Please enter a valid value' // Check if month is an integer
+    ]
+  }
+
+  // Emit the expiry date.
+  @Emit('expiry')
+  private expiryChanged (event:string): string {
```
severinbeauvais, line 64: Please run `npm run lint` before you commit further changes. I think it might catch, and maybe fix, the parameter type spacing.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
JazzarKarim: It's going to look like this:
New Design

What do you think?

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
JazzarKarim: Noted. Going to keep this in mind.

src/mixins/date-mixin.ts
JazzarKarim: Okay I'll copy and paste them there.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
severinbeauvais: That looks tighter to me (nice). Do the errors still display OK?

JazzarKarim: Yes! The field is reactive so the size is going to change based on the length of the error.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
JazzarKarim: I ran lint before committing. Weird that it didn't catch it. I'll add the spacing manually.

src/mixins/date-mixin.ts
JazzarKarim: Good catch! I don't even know why I added that in the first place. I think I was testing something and I forgot to remove it.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
JazzarKarim: Sev I updated the design again. I was checking the UXPin and this is how it looks now. Sorry about that.

newest design

severinbeauvais: That's even better!

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
JazzarKarim: Sev I'm going to keep this as is. But now instead of initializing customMonths as a boolean, it's going to be a string.

severinbeauvais: ok

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,112 @@
+<template id="limited-restoration-panel">
+  <v-radio-group class="mt-0 pt-0 ml-8" v-model="selectMonths" mandatory>
+    <v-radio id="twentyFour-radio" name="selectMonths" label="Two Years" :value=24 />
+    <v-radio id="eighteen-radio" name="selectMonths" label="18 months" :value=18 />
+    <v-radio id="twelve-radio" name="selectMonths" label="12 months" :value=12 />
+    <v-radio id="six-radio" name="selectMonths" label="6 months" :value=6 />
+    <v-row class="ml-0 mt-0">
+      <v-radio id="custom-months" name="selectMonths" value="customMonths" >
+      </v-radio>
```
severinbeauvais, line 9: You can combine these as `<v-radio blah blah blah />`.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,112 @@
...
+      type="number"
+      dense
+      hide-spin-buttons
+      min=1
+      max=24
+      :rules="monthRules"
+      v-model="numberOfMonths"
+      :disabled="!customMonths"
+      @change="onMonthsChanged"
+      filled />
```
severinbeauvais, line 22: 1. Indent continued lines, please.
2. What is class "shrink"?

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,112 @@
...
+  private monthRules: ((data: string) => void)[] = []
+
+  /**
+   * Called when the component is mounted.
+   * Auto select the previous months input of user (if any).
+   * If there is no previous input, auto select 24 months.
+   */
+  mounted (): void {
+    this.setMonthRules()
+    let draftMonths = this.subtractDates(this.currentDate, this.expiryDate)
```
severinbeauvais, line 51: Can be a `const` since you don't reassign it.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,112 @@
...
+    }
+  }
+}
+
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+</style>
```
severinbeauvais, line 112: You can delete this style block since you don't have any scoped styles.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -0,0 +1,112 @@
...
+      dense
+      hide-spin-buttons
+      min=1
+      max=24
+      :rules="monthRules"
+      v-model="numberOfMonths"
+      :disabled="!customMonths"
+      @change="onMonthsChanged"
+      filled />
+      <div class="ml-2 mt-2 text--secondary">month(s)</div>
```
severinbeauvais, line 23: This assumes that Vuetify theme colour "secondary" is assigned to the desired gray. I think it would be better t specify one of our specified colours (gray7 or 9 I think).

src/mixins/date-mixin.ts
```diff
@@ -167,4 +167,45 @@ export default class DateMixin extends Vue {
...
+  /**
+   * Add a number of months to a date and return "YYYY-MM-DD".
+   * Date must be in the format of "YYYY-MM-DD" and months is a number
+   * @example (3, 2023-02-03) -> "2023-05-03"
+   * @example (18, 2023-02-03) -> "2024-08-03"
+   */
+  addMonthsToDate (month: number, date: string): string {
+    if (date === null || date === '') {
```
severinbeauvais, line 178: Or you could do `if (!date)`.

Or you could do `date = date || '2023-01-01'`.

Or you could specify the default in the method signature:
`addMonthsToDate (month: number, date = '2023-01-01'): string {`

src/mixins/date-mixin.ts
```diff
@@ -167,4 +167,45 @@ export default class DateMixin extends Vue {
...
+      date = '2023-01-01'
+    }
+    let temp = this.yyyyMmDdToDate(date)
```
severinbeauvais, line 181: Use const. Also on line 183.

src/mixins/date-mixin.ts
severinbeauvais: Ditto for lines 194 and 197.

src/mixins/date-mixin.ts
severinbeauvais: (Use const as much as possible.)

tests/unit/date-mixin.spec.ts
```diff
@@ -76,4 +76,23 @@ describe('Date Mixin', () => {
...
+  it('returns correct values for subtractDates()', () => {
+    expect(vm.subtractDates('2023-02-03', '2024-08-03')).toBe(18)
+    expect(vm.subtractDates('2023-02-03', '2023-04-03')).toBe(2)
+    expect(vm.subtractDates('2023-01-31', '2023-03-01')).toBe(1)
+    expect(vm.subtractDates('2023-01-31', '2023-03-31')).toBe(2)
+    expect(vm.subtractDates('2023-02-03', '2025-02-03')).toBe(24)
+    expect(vm.subtractDates(null, null)).toBe(24)
+    expect(vm.subtractDates('2023-05-05', null)).toBe(24)
+    expect(vm.subtractDates(null, '2023-04-04')).toBe(3)
+  })
```
severinbeauvais, line 97: Awesome! 👍

ChrisGabel23 (approved):

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
JazzarKarim: Class shrink is used so that the text field's width doesn't take the entire space. 
https://stackoverflow.com/questions/59897603/how-to-change-v-text-field-width-in-vuetify

severinbeauvais: Oh, it's a hidden feature! (actually a flex helper class)

👍

tests/unit/date-mixin.spec.ts
JazzarKarim: Glad to hear that! 😄

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
severinbeauvais: Obsolete comment.

severinbeauvais (approved):

PR #162, opened 2023-02-07
15042 Added some values to some enums

severinbeauvais:
*Issue #:* bcgov/entity#15042

*Description of changes:*
- added some filing codes
- added some filing names
- added some filing types


src/enums/filing-codes.ts
```diff
@@ -1,29 +1,38 @@
 /** Filing codes to send to pay. */
```
severinbeauvais, line 1: I also sorted these enums.

JazzarKarim (approved):

ChrisGabel23 (commented):
I noticed there wasn't a version number change, which you usually do. Was one not needed here?

severinbeauvais:
> I noticed there wasn't a version number change, which you usually do. Was one not needed here?

There is no manual version change when creating/updating a shared resource -- it is done (semi) automatically by the 2 Lerna steps (see Readme).

PR #167, opened 2023-02-17
15289 Updated Approval Type Component (new design)

JazzarKarim:
*Issue #:* /bcgov/entity#15289

*Description of changes:*
- Updated the Approval Type component.
- New Props:
  - The date the Notice of the Application for Restoration was published in the BC Gazette.
  - The date the Application for Restoration was mailed to the company.
- New Emitters:
  - The dates selected in the new date pickers. They correspond to the ones mentioned above.

https://user-images.githubusercontent.com/122301442/219514964-dbf9084d-5b27-4e73-9988-6f114308297b.mp4

---
As per the UXPin:
approval uxpin

---
To use, it's going to look like this:

usage


src/components/approval-type/ApprovalType.vue
```diff
@@ -1,34 +1,70 @@
 <template>
   <div id="approval-type">
     <v-row no-gutters>
-      <v-col cols="12" sm="3" class="pr-4">
+      <v-col cols="12" sm="3" class="pr-4 pt-5">
```
severinbeauvais, line 4: This top padding (or margin) should be in the parent. That way, different parents can add different amounts of whitespace.

In a nutshell: padding and margins should be "tight" around a component.

src/components/approval-type/ApprovalType.vue
```diff
@@ -1,34 +1,70 @@
...
-              filled
-            />
+            <v-expand-transition class="pb-0 mb-0">
+              <v-text-field
+                v-if="approvalTypeSelected === ApprovalTypes.VIA_COURT_ORDER"
+                id="court-order-number-input"
+                v-model="courtOrderNumber"
+                label="Court Order Number"
+                :rules="courtOrderNumRules"
+                :disabled="approvalTypeSelected === ApprovalTypes.VIA_REGISTRAR"
```
severinbeauvais, line 22: Doesn't need to be disabled if it's not rendered (per line 17).

src/components/approval-type/ApprovalType.vue
```diff
@@ -1,34 +1,70 @@
...
           />
+          <v-expand-transition>
+            <v-card flat v-if="approvalTypeSelected === ApprovalTypes.VIA_REGISTRAR">
+              <div class="ml-8 mt-3">
+                <span class="v-label">Enter the date the Notice of the Application for Restoration was published in
+                  <br/>
+                  the BC Gazette:
+                </span>
+                <DatePicker
+                class="mt-2"
```
severinbeauvais, line 43: Indent please.

src/components/approval-type/ApprovalType.vue
```diff
@@ -1,34 +1,70 @@
...
           <v-radio v-if="!isCourtOrderOnly"
-             id="registrar-radio" class="mb-0 pt-2"
-             :label="getRadioText(ApprovalTypes.VIA_REGISTRAR)"
-             :value="ApprovalTypes.VIA_REGISTRAR"
+            id="registrar-radio" class="mb-0 pt-2"
+            :label="getRadioText(ApprovalTypes.VIA_REGISTRAR)"
+            :value="ApprovalTypes.VIA_REGISTRAR"
           />
+          <v-expand-transition>
+            <v-card flat v-if="approvalTypeSelected === ApprovalTypes.VIA_REGISTRAR">
```
severinbeauvais, line 36: Can probably use a div here. V-cards are used for a white block on the gray app background.

Might be able to combine this with line 37.

If you really want a separate block for the "if" condition, change it to a template instead (which doesn't get rendered at all).

src/components/approval-type/ApprovalType.vue
```diff
@@ -1,34 +1,70 @@
...
+              <div class="ml-8 mt-3">
+                <span class="v-label">Enter the date the Notice of the Application for Restoration was published in
+                  <br/>
```
severinbeauvais, line 39: Don't do this. This sentence only wrapped this way because the uxpin was a narrow viewport. Here, let the sentence wrap automatically depending on the viewport width -- the parent div should take care of setting its size.

src/components/approval-type/ApprovalType.vue
```diff
@@ -133,6 +207,44 @@ export default class ApprovalType extends Vue {
...
     return '[error]'
   }
+
+  /**
+   * Emit the date the notice of the application for restoration was published in the BC Gazette.
+   * Validate that a date was successfully selected.
+   * @param event is the date that was selected (if any)
+   */
+  @Emit('emitNoticeDate')
+  private noticeDate (event: string): string {
```
severinbeauvais, line 217: Suggest you rename this. It sounds like a variable instead of an action.

src/components/approval-type/ApprovalType.vue
```diff
@@ -64,53 +105,86 @@ export default class ApprovalType extends Vue {
   /** Show only the court order option; remove via registrar option. */
   @Prop({ default: false }) readonly isCourtOrderOnly!: boolean
 
+  /** Draft notice date. */
+  @Prop({ default: '' }) readonly notice!: string
```
severinbeauvais, line 109: I recommend you name this more explicitly.

src/components/approval-type/ApprovalType.vue
```diff
@@ -64,53 +105,86 @@ export default class ApprovalType extends Vue {
...
-    this.$emit('emitValid', status)
-    return status
+    if (this.approvalTypeSelected === this.ApprovalTypes.VIA_COURT_ORDER) {
+      let status = this.$refs.courtNumRef.validate()
+      this.$emit('emitValid', status)
+      return status
+    } else if (this.approvalTypeSelected === this.ApprovalTypes.VIA_REGISTRAR) {
+      // Emit true (valid) if both dates were selected. Emit false (invalid) if at least one was empty.
+      let status = (!!this.noticeDateText && !!this.applicationDateText)
+      this.$emit('emitValid', status)
```
severinbeauvais, line 149: I know you're just following what was already here, but we typically use a `@Emit`-decorated method to emit events. (Consistency = easier to understand.)

Also, events probably shouldn't be named "emitSomething". In the case of validity, you can just emit a "valid" event.

On the other hand, if you're emitted an update to a prop (eg, notice date), then you can do:
```
@Emit('update:noticeDate')
private emitNoticeDate (noticeDate: string): string {
   return noticeDate
}
```
or
```
@Emit('update:noticeDate')
private emitNoticeDate (noticeDate: string): void {}
```

src/components/approval-type/ApprovalType.vue
```diff
@@ -64,53 +105,86 @@ export default class ApprovalType extends Vue {
...
+      return status
+    } else {
+      // Default state. Emit false (invalid).
+      return false
+    }
+  }
+
+  /** Validations rules for the date picker fields. */
+  private setDateRules (): void {
```
severinbeauvais, line 158: This is called from mounted(), which is called just after the component is rendered. So there's only a very short timespan during which the rules are not set. Why not just set them always (ie, when you declare them)?

src/components/approval-type/ApprovalType.vue
```diff
@@ -64,53 +105,86 @@ export default class ApprovalType extends Vue {
...
+  private datePickerRules = []
+  private noticeDateText = ''
+  private applicationDateText = ''
 
   /** Called when component is mounted. */
   mounted (): void {
+    this.setDateRules()
+    this.setCourtNumberRules()
     // Copy props to mutable properties
     if (this.draftApprovedByRegistrar) {
```
severinbeauvais, line 128: This only get run once, very early on -- possibly before the draft is fetched and parsed to store -- meaning any future update to the props is not handled by the component.

One way to handle this is to watch the props and set your local properties then.

src/components/approval-type/ApprovalType.vue
```diff
@@ -64,53 +105,86 @@ export default class ApprovalType extends Vue {
...
+  // Emit the approval type (radio button selected).
   @Emit('emitRadioButtonChange')
   private radioButtonChanged (event: string): void {
     if (event === this.ApprovalTypes.VIA_REGISTRAR) {
       Vue.set(this, 'courtOrderNumber', '')
       Vue.set(this, 'courtOrderNumRules', [])
-    }
-    if (event === this.ApprovalTypes.VIA_COURT_ORDER) {
+    } else if (event === this.ApprovalTypes.VIA_COURT_ORDER) {
       this.setCourtNumberRules()
```
severinbeauvais, line 175: First, this was already called in mounted().

Second, it would not need to get called just in this case (VIA_COURT_ORDER) since there is no parallel call to unset the rules. In order words, they're just always declared.

So, I recommend just declaring them statically (like the date rules per my other comment).

tests/unit/ApprovalType.spec.ts
```diff
@@ -16,12 +16,16 @@ localVue.use(VueRouter)
 function createDefaultComponent (
   draftCourtOrderNumber: string = '',
   draftApprovedByRegistrar: boolean = false,
+  notice: string = '',
+  application: string = '',
   filingType: string = 'restoration'
 ): Wrapper<ApprovalType> {
   return mount(ApprovalType, {
     propsData: {
       draftCourtOrderNumber,
```
severinbeauvais, line 25: I know they're the draft values the first time, but they're coming from the store, right? And you don't know if the store will change (and thus these props need to update things). So I recommend renaming these to simply "courtOrderNumber" and "approvedByRegistrar" (which would also line up nicely with the update events coming back from the component.

tests/unit/ApprovalType.spec.ts
```diff
@@ -61,10 +65,13 @@ describe('Initialize ApprovalType component', () => {
 
   it('component emits events when court order number entered', async () => {
     const wrapper: Wrapper<ApprovalType> = createDefaultComponent()
+    // Select court order radio to show the court order number text-field
+    const radio = wrapper.find('#court-order-radio')
+    radio.setChecked(true)
+    await Vue.nextTick()
```
severinbeauvais, line 71: You _may_ be able to await the setChecked call and remove this await nextTick.

tests/unit/ApprovalType.spec.ts
```diff
@@ -61,10 +65,13 @@ describe('Initialize ApprovalType component', () => {
...
     // Input text into text-field
     const input = wrapper.find('#court-order-number-input')
     input.setValue('89123456')
```
severinbeauvais, line 74: I think you should await setValue. I think it's async, and you may need to wait for reactivity to complete, including validation and events being emitted.

This unit test _may_ seem to be working because you're getting the expected events below, but it's possible those are emissions from a previous event -- before popping the last event, you could verify that the length of the emitted events is 1 greater than before.

severinbeauvais (commented):
I love the little video you made, showing this component behaving correctly. However, I'd like a bit more development on the code. Thanks.

PS You don't need to update the story with other examples? Maybe an example with "no" data and another one with "all" data?

jonathan-longe (commented):
Good work @JazzarKarim

jonathan-longe:
I too liked the video you made.  It makes it easy to see how the shared component works.

src/components/approval-type/ApprovalType.vue
JazzarKarim: Can I only change the padding for this label in the parent? This top padding is used to make the "Approval Type" label on the same line as the first radio button. If I don't have it here, how can I only change it's padding in the parent?

JazzarKarim: This is what it would look like:
without pt

JazzarKarim: Actually, I have a better idea. I'll just remove this entire label and have it in the parent. So the shared component doesn't include this label. What do you think?

severinbeauvais:
> What do you think about exploring the use of named v-models instead of having a prop and emitter for each variable? See: https://vuejs.org/guide/components/v-model.html#multiple-v-model-bindings The benefits of using v-model is that the values are still reactive after the component is mounted. We still have to "wire-up" each field individually but it's better (I think) than having to wire-up a prop and emitter for each field.

This looks good, but what if our component needs to change its local property? It came in as a prop so it's readonly. Doesn't this mean we have to make a local copy to work on?

On the plus side, it looks like a prop named "x" and an event named "update:x" can be used a single thing (named v-model) from the parent's point of view.

src/components/approval-type/ApprovalType.vue
severinbeauvais: It would best for this component to be consistent with the other components. I think that means the label should be in the component.

So what's adding the extra whitespace above the radio buttons? Inspect the elements and tweak the offending one. Let me know if I can help.

JazzarKarim:
> I too liked the video you made. It makes it easy to see how the shared component works.

Thank you Jonathan! 😃

JazzarKarim:
> I love the little video you made, showing this component behaving correctly. However, I'd like a bit more development on the code. Thanks.
> 
> PS You don't need to update the story with other examples? Maybe an example with "no" data and another one with "all" data?

Thanks Sev! I've updated. Let me know what you think.

src/components/approval-type/ApprovalType.vue
JazzarKarim: fixed padding issue
Fixed! No more extra whitespace above the radio buttons.

src/components/approval-type/ApprovalType.vue
```diff
@@ -1,10 +1,10 @@
 <template>
   <div id="approval-type">
     <v-row no-gutters>
-      <v-col cols="12" sm="3" class="pr-4 pt-5">
+      <v-col cols="12" sm="3" class="pr-4">
         <label id="approval-type-label">Approval Type</label>
       </v-col>
-      <v-col cols="12" sm="9">
+      <v-col cols="12" sm="9" class="mt-n4">
```
severinbeauvais, line 7: Yes, this is fine if this v-col needs to be pulled up. Possibly the v-radio-group had some built-in top whitespace that could have been removed instead but this is fine.

src/components/approval-type/ApprovalType.vue
```diff
@@ -53,78 +92,100 @@ export default class ApprovalType extends Vue {
...
   private valid = false
   private approvalTypeSelected = ''
+  private noticeDateText = ''
+  private applicationDateText = ''
+
+  // Date Picker Rules
+  private datePickerRules = [(v: string) => !!v || 'Date is required']
+
+  // Text Field Rules
+  private courtOrderNumRules = [
```
severinbeauvais, line 123: Not serious enough to change, but these don't have to be private -- external code could read these without causing an issue or seeing anything secret -- but they should be readonly so code doesn't try to reassign these.

src/components/approval-type/ApprovalType.vue
```diff
@@ -141,64 +147,45 @@ export default class ApprovalType extends Vue {
...
   }
 
   // Emit the approval type (radio button selected).
-  @Emit('emitRadioButtonChange')
+  @Emit('radioButtonChange')
   private radioButtonChanged (event: string): void {
     if (event === this.ApprovalTypes.VIA_REGISTRAR) {
-      Vue.set(this, 'courtOrderNumber', '')
+      Vue.set(this, 'courtOrderNumberText', '')
       Vue.set(this, 'courtOrderNumRules', [])
```
severinbeauvais, line 172: I think neither of these Vue.set() is needed because reactivity will work fine if you just reassign your local variables to the new values.

As I understand it, Vue.set() is used when the an item/property is added to an array/object. This is because the local variable still points to the same object, so reactivity needs to be triggered.

Another way to handle to handle a new item/property to an existing array/object this is to reassign the local variable to a NEW object.

Anyway, please check if this code works with a simple assignment instead of Vue.set, and if so then simply it.

severinbeauvais, line 172: I think you shouldn't reset the court order num rules.

src/components/approval-type/ApprovalType.vue
severinbeauvais: I still think this component has a issue where its local properties are not updated if the props change.

tests/unit/ApprovalType.spec.ts
```diff
@@ -61,14 +65,17 @@ describe('Initialize ApprovalType component', () => {
...
     const input = wrapper.find('#court-order-number-input')
-    input.setValue('89123456')
-    await Vue.nextTick()
+    await input.setValue('89123456')
 
-    expect(wrapper.emitted('emitCourtNumberChange').pop()[0]).toEqual('89123456')
-    expect(wrapper.emitted('emitRadioButtonChange').pop()[0]).toEqual('VIA COURT ORDER')
-    expect(wrapper.emitted('emitValid').pop()[0]).toEqual(true)
+    expect(wrapper.emitted('courtNumberChange')[0][0]).toEqual('89123456')
+    expect(wrapper.emitted('radioButtonChange').pop()[0]).toEqual('VIA COURT ORDER')
```
severinbeauvais, line 77: Why do you look at only the last event (using pop())? As this is a fixed test case, you should know how many times this event was emitted before you checked the radio button and set the value, and now you can verify any new expected events.

Ditto line 78.

severinbeauvais (approved):
I think this is good enough to merge. Do you agree?

Then you can work on integrating this component in Create UI and Edit UI.

JazzarKarim:
> I think this is good enough to merge. Do you agree?
> 
> Then you can work on integrating this component in Create UI and Edit UI.

Agreed Sev. I updated according to your latest comments. Shall I go ahead and merge? Also, I already have a working prototype in Create UI as I showed you. I just need now to import and I'll be creating a draft PR for that.

PR #171, opened 2023-02-23
15183 - Added Genesys Web Message Component

JazzarKarim:
*Issue #:* /bcgov/entity#15183

*Description of changes:*
- Added the new Genesys Web Chat Component:

new chat


src/components/genesys-web-message/GenesysWebMessage.vue
```diff
@@ -0,0 +1,57 @@
...
+  @Prop({ default: '0' }) readonly deploymentKey!: string
+
+  mounted (): void {
+    if (this.deploymentKey !== '0') {
+      this.initWebMsg(
+        window,
+        'Genesys',
+        'https://apps.cac1.pure.cloud/genesys-bootstrap/genesys.min.js',
+        {
+          environment: 'cac1',
```
severinbeauvais, line 26: I think the URL and environment would use props, too -- for more future flexibility.

JazzarKarim: I actually had them there. I removed them after our meeting with them. I'll re-add them then!

src/components/genesys-web-message/GenesysWebMessage.vue
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  /**
+   * From Maximus.
+   * @param {Window} g the global object
+   * @param {string} e the 'Genesys' string
+   * @param {string} n the link to the Genesys third-party library
+   * @param {string} es the object with the environment and deployment IDs
+   * @param {HTMLScriptElement} ys the script to be created
```
severinbeauvais, line 40: The element to be created. This element is added as a child to the document head.

It happens to have a JS library as its source, but that's separate from this param.

src/components/genesys-web-message/GenesysWebMessage.vue
```diff
@@ -0,0 +1,57 @@
...
+          environment: 'cac1',
+          deploymentId: this.deploymentKey
+        },
+        null
+      )
+    }
+  }
+
+  /**
+   * From Maximus.
```
severinbeauvais, line 35: Maybe expand on this?

src/components/genesys-web-message/GenesysWebMessage.vue
```diff
@@ -0,0 +1,57 @@
...
+</template>
+
+<script lang="ts">
+// Libraries
+import Vue from 'vue'
+import { Component, Prop } from 'vue-property-decorator'
+
+/**
+ * This component creates the button to generate a chat with Service BC Helpdesk.
+ * This component cannot be unit tested because it heavily relies on a third-party library.
```
severinbeauvais, line 12: I think this second comment should be in a README.md file in this folder, along with info on how to use this component (ie, a code snippet from App.vue or something). And also some info on the props to provide.

src/components/genesys-web-message/GenesysWebMessage.vue
severinbeauvais: Also, I see from Singaravel's signature that it's "MAXIMUS Canada".

src/components/genesys-web-message/GenesysWebMessage.vue
```diff
@@ -8,22 +8,27 @@ import Vue from 'vue'
...
 @Component({})
 export default class GenesysWebMessage extends Vue {
+  @Prop({ default: '' }) readonly genesysURL!: string
+
+  @Prop({ default: '0' }) readonly environmentKey!: string
+
   @Prop({ default: '0' }) readonly deploymentKey!: string
 
   mounted (): void {
     if (this.deploymentKey !== '0') {
```
severinbeauvais, line 25: Should this check `genesysURL` and `environmentKey` as well?

JazzarKarim: I actually wanted to put the default of these two their respective values (the url and environments provided). What do you think?

JazzarKarim: So in the UI, if these two props weren't passed, the component would still work.

src/components/genesys-web-message/README.md
```diff
@@ -0,0 +1,24 @@
...
+- The genesys URL which is the URL provided by Maximus to their Third-Party Library.
+- The Environment Key which is provided by Maximus.
+- The Deployment Key which is also provided by Maximus depending on which UI the component is going to be on.
+
+How To Use
+-
+- Insert the component into the preferred UI with the correct properties. For example:
+```
+<GenesysWebMessage
+    genesysURL="https://apps.cac1.pure.cloud/genesys-bootstrap/genesys.min.js"
```
severinbeauvais, line 18: Add a v-if example, eg.
```
v-if="showWebMessage"
```

src/components/genesys-web-message/README.md
```diff
@@ -0,0 +1,24 @@
...
+    environmentKey="cac1"
+    deploymentKey="5c012aaa-1f17-4d0b-8da2-bea9c1133d5d"
```
severinbeauvais, line 20: Don't use real values here.

src/components/genesys-web-message/GenesysWebMessage.stories.ts
```diff
@@ -0,0 +1,27 @@
...
+  genesysURL: 'https://apps.cac1.pure.cloud/genesys-bootstrap/genesys.min.js',
+  environmentKey: 'cac1',
+  deploymentKey: '5c012aaa-1f17-4d0b-8da2-bea9c1133d5d'
+}
+
+export const SPGP = Template.bind({})
+SPGP.args = {
+  genesysURL: 'https://apps.cac1.pure.cloud/genesys-bootstrap/genesys.min.js',
+  environmentKey: 'cac1',
+  deploymentKey: '42ed05f4-d545-436b-ba2c-94b66ed3396f'
```
severinbeauvais, line 26: The values here are OK since they are for dev or test environments. That means these stories will work.

src/components/genesys-web-message/GenesysWebMessage.vue
severinbeauvais: No, don't make this component work "incorrectly" if the props aren't passed. These props are for a specific UI and environment.

JazzarKarim: I'll add the check then, it's better.

severinbeauvais: You could also simplify this a bit in one of two ways:

1. change `{ default: '0' }` to `{ default: '0' }`, and then you can simply check "if (environmentKey)"

2. change `{ default: '0' }` to `{ required: true }`, which forces the parent to provide the props, but then you'd still have to check that they're not "0"

src/components/genesys-web-message/GenesysWebMessage.vue
```diff
@@ -15,14 +15,14 @@ import { Component, Prop } from 'vue-property-decorator'
  */
 @Component({})
 export default class GenesysWebMessage extends Vue {
-  @Prop({ default: '' }) readonly genesysURL!: string
+  @Prop({ default: null }) readonly genesysURL!: string
 
-  @Prop({ default: '0' }) readonly environmentKey!: string
+  @Prop({ default: null }) readonly environmentKey!: string
 
   @Prop({ default: '0' }) readonly deploymentKey!: string
```
severinbeauvais, line 22: You can default this one to null also, thus simplifying the check on line 25.

src/components/genesys-web-message/README.md
severinbeauvais: Thanks. Looks good.

src/components/genesys-web-message/README.md
severinbeauvais: I don't understand the if variable you added.

src/components/genesys-web-message/README.md
```diff
@@ -0,0 +1,25 @@
+# Genesys Web Message
+
+- This component creates the button to generate a chat with Service BC Helpdesk.
+
+- This component cannot be unit tested because it heavily relies on a third-party library.
+
+Props
+-
+- The genesys URL which is the URL provided by Maximus to their Third-Party Library.
```
severinbeauvais, line 9: Change Maximus to MAXIMUS Canada.

src/components/genesys-web-message/GenesysWebMessage.vue
severinbeauvais: Oh shoot - this is good if the parent provides a null value, but actually that key is a string from 1Password / config file, and we set it to a value of "0" to disable things, so you were absolutely correct to check it for `!== '0'`.

As for the URL and the environment props, well, I'm not going to set those to "0" in 1Password to disable this feature (when setting the deployment key to 0 is enough)... But I can't set those to "" either as 1Password/CD scripts will see that as a key mismatch between environments... So... 

Maybe the best solution is a bit of both:
1. default the props to null and ensure they are not falsy (as above)
2. verify that the deployment key is not "0"

Sorry, and please and thanks.

severinbeauvais (approved):
I think this will be good after that last key fix.

Please poke someone else to review this before merging.

src/components/genesys-web-message/README.md
severinbeauvais: How about a v-if variable that relates to showing this component?

ozamani9gh (approved):
the code i understand from this pull request looks good to me.

severinbeauvais (changes requested):
This component needs to clear the chat history (remove local storage variable) when it is destroyed.

See https://vuejs.org/guide/essentials/lifecycle.html

src/components/genesys-web-message/GenesysWebMessage.vue
severinbeauvais: Looks good.

src/components/genesys-web-message/GenesysWebMessage.vue
```diff
@@ -36,6 +36,10 @@ export default class GenesysWebMessage extends Vue {
     }
   }
 
+  destroyed (): void {
+    localStorage.removeItem('_actmu')
+  }
+
```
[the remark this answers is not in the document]
JazzarKarim: Yes I did! Working fine. Alright will do.

severinbeauvais (approved):
LGTM

GTG after new comment.

PR #173, opened 2023-03-01
Updated the limited restoration radio panel to respond to UX notes

JazzarKarim:

*Description of changes:*
- Updated the limited restoration radio panel in response to the UX assurance notes from Ethan.
- New Design:
new default
new error

For reference: https://github.com/bcgov/entity/issues/14797#issuecomment-1426244089

### DIFF
--- a/src/components/limited-restoration-panel/LimitedRestorationPanel.vue
+++ b/src/components/limited-restoration-panel/LimitedRestorationPanel.vue
@@ -1,14 +1,16 @@
 <template id="limited-restoration-panel">
   <v-radio-group class="mt-0 pt-0 ml-8" v-model="selectMonths" mandatory>
-    <v-radio id="twentyFour-radio" name="selectMonths" label="Two Years" :value=24 />
-    <v-radio id="eighteen-radio" name="selectMonths" label="18 months" :value=18 />
-    <v-radio id="twelve-radio" name="selectMonths" label="12 months" :value=12 />
-    <v-radio id="six-radio" name="selectMonths" label="6 months" :value=6 />
-    <v-row class="ml-0 mt-0">
+    <v-radio class="radio-button pt-2" id="twentyFour-radio" name="selectMonths" label="2 years" :value=24 />
+    <v-radio class="radio-button pt-2" id="eighteen-radio" name="selectMonths" label="18 months" :value=18 />
+    <v-radio class="radio-button pt-2" id="twelve-radio" name="selectMonths" label="12 months" :value=12 />
+    <v-radio class="radio-button pt-2" id="six-radio" name="selectMonths" label="6 months" :value=6 />
+    <v-row class="ml-0 mt-0 radio-button pt-2">
       <v-radio id="custom-months" name="selectMonths" value="customMonths" />
-      <v-text-field
-        class="shrink"
-        hide-details="auto"
+      <v-form ref="monthsRef">
+        <v-text-field
+        id="months-text-field"
+        class="number-months-field"
+        style="width:3.5rem;"
         type="number"
         dense
         hide-spin-buttons
@@ -19,7 +21,8 @@
         :disabled="!customMonths"
         @change="onMonthsChanged"
         filled
-      />
+        />
+      </v-form>
       <div class="ml-2 mt-2 month-text">month(s)</div>
     </v-row>
   </v-radio-group>
@@ -28,17 +31,25 @@
 <script lang="ts">
 import { Component, Emit, Mixins, Prop, Watch } from 'vue-property-decorator'
 import { DateMixin } from '@/mixins' // NB: local mixin (StoryBook can't find it otherwise)
+import { FormIF } from '@bcrs-shared-components/interfaces'
 
 @Component({})
 export default class LimitedRestorationPanel extends Mixins(DateMixin) {
+  // Refs
+  $refs!: Vue['$refs'] & {
+    monthsRef: FormIF
+  }
+
   @Prop({ default: '' }) readonly currentDate!: string
 
   @Prop({ default: '' }) readonly expiryDate!: string
 
+  @Prop({ default: 24 }) readonly maxNumberOfMonths!: number
+
   // Local properties
   private customMonths = ''
   private selectMonths = ''
-  private numberOfMonths = 1
+  private numberOfMonths: number = null
   private monthRules: ((data: string) => void)[] = []
 
   /**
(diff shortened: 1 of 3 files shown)


src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -1,25 +1,26 @@
...
+    <v-radio class="radio-button pt-2" id="twelve-radio" name="selectMonths" label="12 months" :value=12 />
+    <v-radio class="radio-button pt-2" id="six-radio" name="selectMonths" label="6 months" :value=6 />
+    <v-row class="ml-0 mt-0 radio-button pt-2">
       <v-radio id="custom-months" name="selectMonths" value="customMonths" />
-      <v-text-field
-        class="shrink"
-        hide-details="auto"
+      <v-form ref="monthsRef">
+        <v-text-field
+        id="months-text-field"
```
severinbeauvais, line 11: Please indent these lines. Leave the `/>` aligned with the `<v-text-field`.

Ideally our IDEs (or linter) would do this automatically. Anyone care to look into a "prettier" extension?

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -61,31 +70,37 @@ export default class LimitedRestorationPanel extends Mixins(DateMixin) {
...
-      (v:string) => (v > 0 && v <= 24) || 'This field must be between 1 to 24', // Keep month between 1 and 24
-      (v:string) => (v % 1 === 0) || 'Please enter a valid value' // Check if month is an integer
+      (v:string) => !!v || 'Must be between 1 to ' + this.maxNumberOfMonths,
+      (v:string) => (v > 0 && v <= this.maxNumberOfMonths) || 'Must be between 1 to ' + this.maxNumberOfMonths,
+      (v:string) => (v % 1 === 0) || 'Must be between 1 to ' + this.maxNumberOfMonths // Check if month is an integer
     ]
   }
 
+  /** Triggers the form validation. */
+  public validate (): void {
```
severinbeauvais, line 80: This suggests you want the parent to call this function. I think it would be better to watch a `validate` prop. Eg,
https://github.com/bcgov/bcrs-shared-components/blob/78926fd0021ff0e782d8d4d302ec968e1134daf5/src/components/certify/Certify.vue#L147

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -97,19 +112,31 @@ export default class LimitedRestorationPanel extends Mixins(DateMixin) {
...
 </script>
 
 <style lang="scss" scoped>
 @import '@/assets/styles/theme.scss';
-
 .month-text {
   color: $gray7;
 }
+
+::v-deep .number-months-field{
```
severinbeauvais, line 135: Add a comment explaining what this is doing (or why).

Also, need a space before `{`.

tests/unit/LimitedRestorationPanel.spec.ts
```diff
@@ -15,12 +15,14 @@ localVue.use(VueRouter)
  */
 function createDefaultComponent (
   currentDate: string = '2023-02-03',
-  expiryDate: string = null
+  expiryDate: string = null,
+  maxNumberOfMonths: number = 24
```
severinbeauvais, line 19: Maybe you shouldn't default this, so you can test the component's default. Does this work if you default to null here? Then you can add a test to verify a different max.

JazzarKarim: It doesn't work! I believe the default value of 24 is not being set properly. If I set it manually like this, it works just fine. Shall I leave it as is or try to work around it?

severinbeauvais: I think the problem is "null" is being passed in, which is different than the prop not being passed at all.

Leave it.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
JazzarKarim: I don't get this. I'm calling this function whenever the number of months entered is changed. Each time it's changed, I check if the number is valid or not by calling this function. Hence, why would I watch a validate `prop`?

JazzarKarim: It's basically like this here:
https://github.com/bcgov/bcrs-shared-components/blob/78926fd0021ff0e782d8d4d302ec968e1134daf5/src/components/approval-type/ApprovalType.vue#L149

severinbeauvais: Maybe I misunderstood. I saw that it's public (for external use, like a parent asking a child to do something) instead of private (for internal use only). I assumed it would be called to validate the current form (like some other components do).

It seems that you only call it from 1 place. I think it would be less confusing if you moved its contents into `onMonthsChanged()`.

severinbeauvais: > It's basically like this here:

Yes, I see that now. It's not used as I expected (sorry).

JazzarKarim: Great idea actually. I will do that.

severinbeauvais: Thanks.

To explain where I was coming from: some component are told when to validate the form (eg, when step-wise or filing-wise validation is enabled). One way that could be done is with a public function (as above) that the parent could call, like this: `this.$refs.mySubcomponent.validate()`, but this isn't good practice (except for form components, where we often call validate or reset on them). Another way is to have a prop, `validate`, that the component watches and reacts to. This may not be good practice either, but many of our components do that so at least it's consistent :)

JazzarKarim: Hmm okay! Got it. I'll keep that in mind moving forward.

severinbeauvais (approved):

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -1,25 +1,24 @@
...
-        :disabled="!customMonths"
-        @change="onMonthsChanged"
-        filled
-      />
+    <v-radio class="radio-button pt-2" id="twentyFour-radio" name="selectMonths" label="2 years" :value=24 />
+    <v-radio class="radio-button pt-2" id="eighteen-radio" name="selectMonths" label="18 months" :value=18 />
+    <v-radio class="radio-button pt-2" id="twelve-radio" name="selectMonths" label="12 months" :value=12 />
+    <v-radio class="radio-button pt-2" id="six-radio" name="selectMonths" label="6 months" :value=6 />
+    <v-row class="ml-0 mt-0 radio-button pt-2">
+      <v-radio class="mt-n4" id="custom-months" name="selectMonths" value="customMonths" />
```
severinbeauvais, line 8: Could "mt-n4" be moved to line 7?

JazzarKarim: No, this is done to center only the radio bubble with the text field. I showed it to Ethan yesterday and that's how the UX team wants it!

JazzarKarim: Without the "mt-n4" in v-radio:
without
With the "mt-n4" in v-radio:
with

severinbeauvais: Ah, OK.

ketaki-deodhar (approved):
LGTM

PR #174, opened 2023-03-06
Emit validation on component load

JazzarKarim:

*Description of changes:*
- Limited restoration now validates on load when a custom number of months option is selected.
- https://github.com/bcgov/business-create-ui/pull/522#discussion_r1125416120

### DIFF
--- a/src/components/limited-restoration-panel/LimitedRestorationPanel.vue
+++ b/src/components/limited-restoration-panel/LimitedRestorationPanel.vue
@@ -60,6 +60,10 @@ export default class LimitedRestorationPanel extends Mixins(DateMixin) {
       this.selectMonths = 'customMonths'
       this.numberOfMonths = draftMonths
       this.setCustomMonths(this.numberOfMonths)
+      // Emit validation on load
+      if (this.numberOfMonths > this.maxNumberOfMonths || this.numberOfMonths < 1) {
+        this.monthsValid(false)
+      }
     } else {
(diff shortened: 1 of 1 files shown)


src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -60,6 +60,10 @@ export default class LimitedRestorationPanel extends Mixins(DateMixin) {
       this.selectMonths = 'customMonths'
       this.numberOfMonths = draftMonths
       this.setCustomMonths(this.numberOfMonths)
+      // Emit validation on load
+      if (this.numberOfMonths > this.maxNumberOfMonths || this.numberOfMonths < 1) {
+        this.monthsValid(false)
+      }
```
severinbeauvais, line 66: Yes, I think it's fine for a component to emit validation on load.

One thing to watch for: this assumes the data is current when the component loads. In some cases, the initial data is replaced when, say, a draft is finished loading -- so the component also needs to react to that. I don't know if that's the case here; if it tests OK then you're probably fine.

severinbeauvais (approved):

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
JazzarKarim: I tested it and seems to be working fine! That's a good point though. I'll keep that in mind.

PR #175, opened 2023-03-09
Remove themeing from relationship panel

jonathan-longe:
*Issue #:* /bcgov/entity#15092

Remove padding, margin and background colour from relationship panel.  Here's what it looked like prior to the changes.  These changes make the shared component more flexible because the parent can theme as needed.

image


src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -103,7 +103,6 @@ export default class RelationshipsPanel extends Vue {
 
 .relationships-vcard {
   font-size: 1rem;
-  background-color: rgba(0, 0, 0, 0.06);
```
jonathan-longe, line 106: Removing the background colour

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -1,6 +1,6 @@
 <template>
-  <v-card flat id="relationships-panel" class="pb-4 pl-4">
-    <v-card flat rounded="sm" class="relationships-vcard px-4 mt-3 mr-9 ml-4">
+  <v-card flat id="relationships-panel">
+    <v-card flat rounded="sm" class="relationships-vcard">
```
severinbeauvais, line 3: I just noticed this -- there shouldn't be a v-card in a v-card. Probably the second one should be a div.

src/components/relationships-panel/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bcrs-shared-components/relationships-panel",
-  "version": "1.0.9",
+  "version": "1.0.10",
```
severinbeauvais, line 3: Don't change this -- Lerna does this.

src/components/relationships-panel/RelationshipsPanel.vue
jonathan-longe: Wow, good catch @severinbeauvais  I'll fix it

src/components/relationships-panel/RelationshipsPanel.vue
severinbeauvais, line 106: Removing this means the parent will have to reach into this component (using deep) to set this block's colour. That's generally frowned upon. Instead, add a prop (and default it as needed) to conditionally set this.

JazzarKarim (approved):
Looks good. I won't be adding any spacing for shared components moving forward. Sorry about that Jonathan!

src/components/relationships-panel/package.json
severinbeauvais: PS steps 14 and 15 can only be done by a repo admin (since they affect the main branch directly).

severinbeauvais:
> Looks good. I won't be adding any spacing for shared components moving forward. Sorry about that Jonathan!

"External whitespace" (outer edge of a component) should not exist.

"Internal whitespace" (inside component) is OK.

src/components/relationships-panel/package.json
jonathan-longe: Whoops, thanks for the reminder.  I'll fix it.

JazzarKarim:
> > Looks good. I won't be adding any spacing for shared components moving forward. Sorry about that Jonathan!
> 
> "External whitespace" (outer edge of a component) should not exist.
> 
> "Internal whitespace" (inside component) is OK.

Alright. I'll be planting this in my head so that I won't forget. Thanks Sev.

src/components/relationships-panel/RelationshipsPanel.vue
jonathan-longe: What about using CSS?  Can't I theme the `.relationship-vcard` class from the parent?  Wouldn't that be more flexible?  I'm happy to pass in a prop if that's the way you'd prefer to do it.

severinbeauvais: `:deep()` is a CSS selector used for reaching into a sub-component (possibly Shadow DOM).

I think it's a bad idea to reach into components to style them. What do you think?

severinbeauvais: PS Maybe rename that class now? relationship-wrapper or relationship-content or something

jonathan-longe: I'll go with your suggestion.  Changes coming up in just a sec ...

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -43,6 +43,7 @@ import { RelationshipTypes } from '@bcrs-shared-components/enums'
 export default class RelationshipsPanel extends Vue {
   /** Draft restoration relationships */
   @Prop({ default: () => [] }) readonly draftRelationships!: RelationshipTypes[]
+  @Prop({default: '#fff'}) readonly bgHex!: string
```
severinbeauvais, line 46: Linting didn't catch this? There should a space after { and before }.

severinbeauvais (approved):

PR #176, opened 2023-03-11
Hide immediate validation in RelationShipsPanel

jonathan-longe:
*Issue #:* /bcgov/entity#15092

Hide validation errors when the RelationshipsPanel first loads.  


src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -2,32 +2,45 @@
...
         <v-col cols="4">
-          <v-checkbox id="heir-legal-rep-checkbox" class="mt-0" v-model="selectedRelationships"
-            :value="RelationshipTypes.HEIR_LEGAL_REP" :label="RelationshipTypes.HEIR_LEGAL_REP"
-            :rules="relationshipRules" />
+          <v-checkbox id="heir-legal-rep-checkbox" class="mt-0"
+            v-model="selectedRelationships"
+            :value="RelationshipTypes.HEIR_LEGAL_REP"
+            :label="RelationshipTypes.HEIR_LEGAL_REP"
+            :error="displayErrorState"
+          />
```
jonathan-longe, line 11: Trying a different approach here.  Instead of using the `rules` prop to show / hide validation errors, I'm using the `error` prop to show / hide the validation errors.  The difference is that using rules we add/remove the appropriate rules to show/hide the validation errors.  Using the `error` prop, we no longer change the rules dynamically, instead we show and hide the display of the validation errors.

jonathan-longe, line 11: Trying a different approach here.  Instead of using the `rules` prop to show / hide validation errors, I'm using the `error` prop to show / hide the validation errors.  The difference is that using rules we add/remove the appropriate rules to show/hide the validation errors.  Using the `error` prop, we no longer change the rules dynamically, instead we show and hide the display of the validation errors.

src/components/relationships-panel/RelationshipsPanel.vue
```diff
@@ -58,16 +71,12 @@ export default class RelationshipsPanel extends Vue {
   mounted (): void {
     if (this.draftRelationships.length > 0) {
       this.selectedRelationships.push(...this.draftRelationships)
-    } else {
-      this.selectedRelationships.push() // This is done to validate component immediately (no need to touch)
```
JazzarKarim, line 62: There is no need for this now since the new `error` prop will take care of component validation on load right?

src/components/relationships-panel/RelationshipsPanel.vue
severinbeauvais: OK, thanks for trying something different :)

When does the component report that it's invalid?

I looked at the code below. There's a tiny bit of duplication but it looks so simple I'm happy with it.

tests/unit/RelationshipsPanel.spec.ts
```diff
@@ -2,60 +2,64 @@ import Vue from 'vue'
...
 const localVue = createLocalVue()
-localVue.use(VueRouter)
 
 /**
  * Creates and mounts a blank, un-populated component
  */
-function createDefaultComponent (
-  draftRelationships: string[] = []
-): Wrapper<RelationshipsPanel> {
+function createDefaultComponent (draftRelationships = [], showValidationErrors = false) {
```
severinbeauvais, line 13: I love that the parameters here are the same props passed to the component.

tests/unit/RelationshipsPanel.spec.ts
```diff
@@ -2,60 +2,64 @@ import Vue from 'vue'
...
-    const wrapper: Wrapper<RelationshipsPanel> = createDefaultComponent()
+  it('no relationship selected (draft relationship empty)', () => {
+    const wrapper = createDefaultComponent()
     expect(wrapper.vm.$data.selectedRelationships).toEqual([])
     wrapper.destroy()
   })
 
-  it('component emits events when no relationship is selected (validate)', async () => {
-    const wrapper: Wrapper<RelationshipsPanel> = createDefaultComponent()
+  it('component emits a valid=true event when at least one relationship selected', async () => {
```
severinbeauvais, line 37: "it emits..."

src/components/relationships-panel/RelationshipsPanel.vue
jonathan-longe: @severinbeauvais good question.  The component always reports its current validation status.  By using the `error` prop, I'm suggesting that we toggle on / off is whether the component displays the validation status to the user.

src/components/relationships-panel/RelationshipsPanel.vue
jonathan-longe: Yes, that's right @JazzarKarim the component now is always reporting it's validation status.  What we're toggling on and off via the `error` prop is whether those validation errors are shown to the user.

tests/unit/RelationshipsPanel.spec.ts
```diff
@@ -2,60 +2,64 @@ import Vue from 'vue'
...
   it('loads the component', () => {
-    const wrapper: Wrapper<RelationshipsPanel> = createDefaultComponent()
+    const wrapper = createDefaultComponent()
     expect(wrapper.findComponent(RelationshipsPanel).exists()).toBe(true)
     wrapper.destroy()
   })
 
-  it('no relationship selected (draft relationship empty)', async () => {
-    const wrapper: Wrapper<RelationshipsPanel> = createDefaultComponent()
+  it('no relationship selected (draft relationship empty)', () => {
```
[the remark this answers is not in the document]
jonathan-longe: Good point, I'll fix it.

tests/unit/RelationshipsPanel.spec.ts
jonathan-longe: You bet, I'll fix this too

tests/unit/RelationshipsPanel.spec.ts
```diff
@@ -2,60 +2,64 @@ import Vue from 'vue'
...
+    const wrapper = createDefaultComponent()
     expect(wrapper.vm.$data.selectedRelationships).toEqual([])
     wrapper.destroy()
   })
 
-  it('component emits events when no relationship is selected (validate)', async () => {
-    const wrapper: Wrapper<RelationshipsPanel> = createDefaultComponent()
+  it('component emits a valid=true event when at least one relationship selected', async () => {
+    const wrapper = createDefaultComponent()
+    wrapper.find('#shareholder-checkbox').trigger('click')
```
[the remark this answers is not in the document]
jonathan-longe: Good point, I'll try it.

src/components/relationships-panel/RelationshipsPanel.vue
JazzarKarim: This is actually something that I've had trouble dealing with in this and other components. Amazing idea. Thanks Jonathan 😃

severinbeauvais (approved):

ketaki-deodhar (approved):

tests/unit/RelationshipsPanel.spec.ts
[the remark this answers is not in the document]
jonathan-longe: That was a great suggestion @severinbeauvais Removing `await Vue.nextTick()` and adding await to the `trigger('click')` works beautifully!

jonathan-longe:
Over to you @severinbeauvais  Can you work your npm publishing magic?

PR #181, opened 2023-05-09
16306 - Get rid of array.prototype.move + add in CI and fix unit tests

seeker25:
Update various package + package-locks, small unit test update.

*Issue #:* /bcgov/entity#16306

*Description of changes:*
Get rid of array.prototype.move. This prototype isn't allowed by the more recent versions of pdfjs-dist.

https://github.com/mozilla/pdfjs-dist/blob/5f07d5a4159bb99eee2f6143d1297f03b45bba58/lib/core/worker.js#L95


package.json
```diff
@@ -14,7 +14,7 @@
   "dependencies": {
     "@babel/compat-data": "^7.19.1",
     "@mdi/font": "^5.9.55",
-    "array.prototype.move": "0.0.4",
+    "array-move": "4.0.0",
```
severinbeauvais, line 17: 1. This package file is for Storybook.

2. Why not `^4.0.0` ?

severinbeauvais: I'm thinking you don't need this import ...?

tests/unit/EditShareStructure.spec.ts
```diff
@@ -149,7 +149,7 @@ describe('Edit Share Structure component', () => {
     const wrapper: Wrapper<EditShareStructure> = createComponent(shareClass, -1, 1, null, [])
     await Vue.nextTick()
 
-    wrapper.find(doneButtonSelector).trigger('click')
+    await wrapper.find(doneButtonSelector).trigger('click')
     await Vue.nextTick()
```
severinbeauvais, line 153: Since you're now awaiting the trigger, you may not need to await until the next tick. Ditto below.

seeker25: I tried without nextTick, still didn't work

src/components/share-structure/package.json
```diff
@@ -9,7 +9,7 @@
     "@bcrs-shared-components/confirm-dialog": "^1.2.2",
     "@bcrs-shared-components/enums": "^1.0.38",
     "@bcrs-shared-components/interfaces": "^1.0.60",
-    "array.prototype.move": "0.0.4",
+    "array-move": "4.0.0",
```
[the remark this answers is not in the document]
seeker25: It was ^4.0.0, just left it at 4. hasn't been updated in a while

package.json
seeker25: Need it for the shares component I thought?

severinbeauvais (approved):
LGTM.

See comments and maybe add 1 or 2 other devs that should know about this repo/code.

tests/unit/EditShareStructure.spec.ts
severinbeauvais: OK, sometimes that happens. No worries.

src/components/share-structure/package.json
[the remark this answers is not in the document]
severinbeauvais: But wouldn't we want to automatically pick up fixes or improvements, if any?

package.json
severinbeauvais: So the shares component has its own package file where it depends on it.

THIS package file is for Storybook only (I think).

src/components/share-structure/package.json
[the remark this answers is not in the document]
seeker25: I don't think there are going to be very many fixes - honestly probably should have just incorporated it as a util:

```
export function arrayMoveMutable(array, fromIndex, toIndex) {
	const startIndex = fromIndex < 0 ? array.length + fromIndex : fromIndex;

	if (startIndex >= 0 && startIndex < array.length) {
		const endIndex = toIndex < 0 ? array.length + toIndex : toIndex;

		const [item] = array.splice(fromIndex, 1);
		array.splice(endIndex, 0, item);
	}
}

export function arrayMoveImmutable(array, fromIndex, toIndex) {
	const newArray = [...array];
	arrayMoveMutable(newArray, fromIndex, toIndex);
	return newArray;
}
```

seeker25:
I've included the library in the code, plus fixed some other unit tests. Works locally, not sure if it will work on github actions.

package.json
seeker25: Removed

jest.config.js
```diff
@@ -6,6 +6,17 @@ module.exports = {
       babelConfig: true
     }
   },
+  // Have to do this for LINUX right now as a work around.
+  // I noticed this is done in one of the webpack files as well.
+  // It's possible the @ in the name of the package get confused with the alias somehow.
+  moduleNameMapper: {
```
seeker25, line 12: Had to add this to make it work with linux

src/components/correct-name/CorrectName.vue
```diff
@@ -68,9 +68,9 @@ import { Component, Emit, Prop } from 'vue-property-decorator'
 import { NameRequestIF } from '@bcrs-shared-components/interfaces'
 import { CorrectNameOptions } from '@bcrs-shared-components/enums'
 import { CorpTypeCd } from '@bcrs-shared-components/corp-type-module'
-import CorrectCompanyName from '@bcrs-shared-components/correct-name/CorrectCompanyName.vue'
-import CorrectNameToNumber from '@bcrs-shared-components/correct-name/CorrectNameToNumber.vue'
-import CorrectNameRequest from '@bcrs-shared-components/correct-name/CorrectNameRequest.vue'
+import CorrectCompanyName from './CorrectCompanyName.vue'
```
seeker25, line 71: Changed these to relative imports

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -26,7 +26,7 @@
 
 <script lang="ts">
 import { Component, Emit, Mixins, Prop, Watch } from 'vue-property-decorator'
-import { DateMixin } from '@bcrs-shared-components/mixins'
+import { DateMixin } from '@/mixins' // NB: local mixin (StoryBook can't find it otherwise)
```
seeker25, line 29: Another small fix

src/components/share-structure/array-move.ts
```diff
@@ -0,0 +1,57 @@
+/** array-move version 4.0
```
seeker25, line 1: No more library, except it's in a file now

tests/unit/corp-type-module.spec.ts
```diff
@@ -18,14 +18,14 @@ describe('Corp Type Module', () => {
   })
 
   it('gets corp info object correctly', () => {
-    const obj = GetCorpInfoObject(CorpTypeCd.BC_CORPORATION)
+    const obj = GetCorpInfoObject(CorpTypeCd.BC_COMPANY)
```
seeker25, line 21: `BC_CORPORATION` is commented out

src/components/share-structure/EditShareStructure.vue
```diff
@@ -240,11 +240,11 @@ export default class EditShareStructure extends Mixins(CurrencyLookupMixin) {
         (v: string) => !(this.shareClasses
           .find((s, index) => {
             // Don't apply uniqueness check to self
-            return index !== this.activeIndex && s.name.split(' Shares')[0].toLowerCase() === v.toLowerCase()
+            return index !== this.activeIndex && s.name?.split(' Shares')[0].toLowerCase() === v.toLowerCase()
```
seeker25, line 243: Not the happiest about these for fixes, but I wanted to get the unit tests to pass - if we need to fix these another way, we can create a ticket and merge with the rest of the tests passing.

seeker25:
@severinbeauvais could you take a look at this before merging again? I made a bunch of other changes

src/components/correct-name/CorrectName.vue
severinbeauvais: Good idea. Dunno what I was thinking when I did this.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
severinbeauvais: I don't like this one, as it means the "parent UI" needs to have this exact mixin in its local src/mixins/ folder.

I'm hoping your `moduleNameMapper` change, or something similar for SB, will allow the old code to work in all situations.

src/components/share-structure/EditShareStructure.vue
severinbeauvais, line 243: What problem is this solving?

(I'm wary of conditional chaining because it can hide things.)

src/components/share-structure/EditShareStructure.vue
```diff
@@ -240,11 +240,11 @@ export default class EditShareStructure extends Mixins(CurrencyLookupMixin) {
...
           })
         ) || 'Class name must be unique')
       rules.push(
-        (v: string) => !(v.split(' ').some(r => this.excludedWordsListForClass.includes(r.toLowerCase()))) ||
+        (v: string) => !(v && v.split(' ').some(r => this.excludedWordsListForClass.includes(r.toLowerCase()))) ||
```
severinbeauvais, line 247: Doesn't the rule on line 239 already ensure that v is truthy?

src/components/share-structure/array-move.ts
severinbeauvais: OK.

In the future, I'd prefer to see this in the src/utils/ file (SnakeCase), like the other standalone functions we have. (This repo is not a great example but see https://github.com/bcgov/business-create-ui/tree/main/src/utils.)

severinbeauvais: I'm flexible on this file's location -- as this is a repo of "standalone" components, I do appreciate having this function/file in the present folder.

I'm thinking ahead about "parent UIs" that might import this as a package, or could use it from a (shared, maybe) utils folder.

src/components/share-structure/EditShareStructure.vue
seeker25: It's solving the unit test failing, there's an error regarding calling split on an undefined object

src/components/share-structure/EditShareStructure.vue
seeker25: That's what I thought as well, but put this in here to get rid of the error - tried in all places where split occured

seeker25:
1. I'll look at 
2. No changes required right now - it's just CI

src/components/share-structure/EditShareStructure.vue
seeker25: reverting this back

src/components/share-structure/EditShareStructure.vue
seeker25: Yeah I needed to have the v check.. I think it validates all of the rules and doesn't short circuit

src/components/share-structure/EditShareStructure.vue
severinbeauvais: Yeah, better to fix the unit test than compromise the regular build.

src/components/share-structure/EditShareStructure.vue
severinbeauvais: I believe it evaluates all rules until it finds a failing one (returns string instead of True)...?
https://vuetifyjs.com/en/components/forms/#rules

Better to fix the unit test though.

src/components/share-structure/EditShareStructure.vue
severinbeauvais: Probably `s` exists but is missing the `name` property in the unit test.

src/components/share-structure/EditShareStructure.vue
severinbeauvais: I'm curious what `v` is on line 247 (if it's not falsy).

(If this is burning up your time then I'm OK with this code change.)

src/components/share-structure/EditShareStructure.vue
seeker25: I removed, wasn't a problem, problem was the v's below

src/components/share-structure/EditShareStructure.vue
seeker25: `        this.shareStructure.name = name.substr(0, name.indexOf(' Shares'))` 
probably not including Shares in the name

seeker25: Hard to tell, debugging breakpoints aren't showing the correct line with this project. I'm just leaving it for now.

PR #182, opened 2023-05-19
16226 Refactored Limited Restoration Panel

severinbeauvais:
*Issue #:* bcgov/entity#16226

*Description of changes:*

This component caused build errors in Edit UI with the new packages, so I had to do something. In the end, to fix this component's behaviour in Edit UI, I basically refactored it.

- updated stories
- refactored to handle only the month value + misc fixes
- updated dependencies
- updated unit tests
- updated Storybook


src/components/limited-restoration-panel/LimitedRestorationPanel.stories.ts
```diff
@@ -14,26 +14,21 @@ const Template = (args, { argTypes }) => ({
 
 export const Default = Template.bind({})
 Default.args = {
-  currentDate: '2023-02-03',
-  expiryDate: null
+  months: 0
```
severinbeauvais, line 17: As you can see, this component now takes simply the "months" value property, and it emits an updated months value (and valid value).

This is a big change so when I publish this then I will increase the major release number.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -1,127 +1,147 @@
 <template id="limited-restoration-panel">
```
severinbeauvais, line 1: ESLint did a log of cleanup here.

I did rename the radio button `name` and the text field `v-model`.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -1,127 +1,147 @@
...
   // Refs
   $refs!: Vue['$refs'] & {
     monthsRef: FormIF
   }
 
-  @Prop({ default: '' }) readonly currentDate!: string
-
-  @Prop({ default: '' }) readonly expiryDate!: string
-
+  @Prop({ required: true }) readonly months!: number
```
severinbeauvais, line 73: I leave it to the parent app to decide what to default this, as needed.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -1,127 +1,147 @@
...
+
+  /**
+   * The validation rules for months input.
+   */
+  get monthsRules (): Array<(v: string) => boolean | string> {
+    return [
+      (v) => (+v > 0 && +v <= this.maxNumberOfMonths) || 'Must be between 1 and ' + this.maxNumberOfMonths,
+      (v) => (+v % 1 === 0) || 'Must be a whole number'
+    ]
+  }
```
severinbeauvais, line 88: This was previously set from the `mounted()` lifecycle event, and only ever had one value, so I moved it here. Note that it's a getter (computed) instead of a static array because it uses a variable internally.

severinbeauvais:
Unit tests:

image

severinbeauvais:
Storybook:

image

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -1,127 +1,148 @@
...
 <script lang="ts">
-import { Component, Emit, Mixins, Prop, Watch } from 'vue-property-decorator'
-import { DateMixin } from '@/mixins' // NB: local mixin (StoryBook can't find it otherwise)
+import Vue from 'vue'
+import { Component, Emit, Prop, Watch } from 'vue-property-decorator'
 import { FormIF } from '@bcrs-shared-components/interfaces'
 
 @Component({})
-export default class LimitedRestorationPanel extends Mixins(DateMixin) {
+export default class LimitedRestorationPanel extends Vue {
```
JazzarKarim, line 67: Why did you change this back to extend Vue Sev? In the UIs, we were moving towards extending Mixins. Why is it not the case here?

severinbeauvais: This component no longer needs the mixin :)

JazzarKarim: Oh OK. My bad. Resolving!

severinbeauvais: No, I'm glad you're looking for such things and questioning them 😃

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -1,127 +1,148 @@
...
-    if (draftMonths !== 24 && draftMonths !== 18 && draftMonths !== 12 && draftMonths !== 6) {
-      this.selectMonths = 'customMonths'
-      this.numberOfMonths = draftMonths
-      this.setCustomMonths(this.numberOfMonths)
-      // Emit validation on load
-      if (this.numberOfMonths > this.maxNumberOfMonths || this.numberOfMonths < 1) {
-        this.monthsValid(false)
-      }
+    if ([24, 18, 12, 6].includes(this.months)) {
+      this.radioValue = this.months.toString()
```
JazzarKarim, line 96: I really really like this. We would have to keep in mind to change the panel in Create and Edit accordingly since the implementation changed. We're going to have to create tickets for those right?

severinbeauvais: I'm updating this component specifically for Edit UI (part of my ticket).

Yes, if Create UI has build errors in this component then it will need to import the latest component (this one). In the meantime, it can keep importing the previous version of this component.

JazzarKarim: Alright, noted.

src/components/limited-restoration-panel/LimitedRestorationPanel.stories.ts
severinbeauvais: Basically I changed this component to only care about the expiry months -- a simpler concern, and the parent app can handle more complicated logic (or in this case, simpler -- you'll see when you review my Edit UI changes).

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
severinbeauvais: > Never mind, I just saw that you've made the change for Edit UI in the cleanup PR! 😃

I have it working but not quite finished!

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -1,127 +1,148 @@
...
+          filled
           hide-spin-buttons
-          :rules="monthRules"
-          v-model="numberOfMonths"
-          :disabled="!customMonths"
-          @change="onMonthsChanged"
-          filled />
+          :rules="monthsRules"
+          :disabled="(radioValue !== 'customMonths')"
+          @input="onMonthsInput($event)"
```
JazzarKarim, line 53: What's the difference between `@change` and `@input`?

severinbeauvais: `@change` fires when the value has changed (been accepted by the component).

`@input` fires when anything in the input has changed, such as typing in "1" or backspacing over it. This allows me to pass this component's "real-time" validity to the parent, since there's no Done button.

JazzarKarim: I wish I knew this earlier 😭 I've struggled with real-time validity on various occasions. Alright, noted. Thanks Sev!

severinbeauvais: For the record, native HTML controls have a ton of events that you can research (Google).

This is a Vuetify component and I didn't find a ton of good info on events, but I played with the documented ones and figured out what they did. (Ref: https://wotamann.gitbook.io/vuetify-form-base/schema/events)

severinbeauvais: I just found this but it doesn't change my code: https://vuetifyjs.com/en/api/v-text-field/#props-validation-value

Here's all the docs have to say about text field events: https://vuetifyjs.com/en/api/v-text-field/#events

I guess you're supposed to guess that text fields inherit base events from the form 🤷‍♂️  😒

JazzarKarim: One thing that I've noticed is that when it comes to Vuetify (or Vue in general), there isn't much information out there as React or Native HTML stuff for example. I would have to dig deep in the docs to find an answer to whatever that I'm trying to achieve. 

This here is a perfect example of that. Thanks for the Refs Sev. Also, I'll be keeping this in mind so that I won't have to guess next time 😟

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -1,127 +1,148 @@
...
-  private monthRules: ((data: string) => void)[] = []
+  radioValue = '' // v-model for radio buttons
+  inputValue = '' // v-model for months text field
+
+  /**
+   * The validation rules for months input.
+   */
+  get monthsRules (): Array<(v: string) => boolean | string> {
+    return [
+      (v) => (+v > 0 && +v <= this.maxNumberOfMonths) || 'Must be between 1 and ' + this.maxNumberOfMonths,
```
JazzarKarim, line 85: What's the `+` doing here Sev?

JazzarKarim (approved):
LGTM! 👍

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
severinbeauvais: Converting v (a string from the text field component) to a number.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus

JazzarKarim: Oh OK. Sorry about this question. I'm used to the native function `Number()`. Thanks!

severinbeauvais: I use whichever one gives me easier to read code. On line 128 I use Number() because the unary operator didn't stand out or looked like a mistake. Here it's much more clear this way IMHO.

PR #183, opened 2023-05-29
Added Continuation Out Properties

JazzarKarim:
*Issue #:* /bcgov/entity#16116

*Description of changes:*
- Added the properties/attributes needed to do Continuation Out filings


src/components/share-structure/EditShareStructure.vue
```diff
@@ -244,7 +244,7 @@ export default class EditShareStructure extends Mixins(CurrencyLookupMixin) {
           })
         ) || 'Class name must be unique')
       rules.push(
-        (v: string) => !(v.split(' ').some(r => this.excludedWordsListForClass.includes(r.toLowerCase()))) ||
+        (v: string) => !(v?.split(' ').some(r => this.excludedWordsListForClass.includes(r.toLowerCase()))) ||
```
JazzarKarim, line 247: There were two broken unit tests not related to this ticket:
fixed unit tests

This line and line 260 fixes these two tests.

severinbeauvais: Why is this needed, since line 239 should validate that the value exists?

src/components/share-structure/EditShareStructure.vue
```diff
@@ -257,7 +257,7 @@ export default class EditShareStructure extends Mixins(CurrencyLookupMixin) {
           })
         ) || 'Series name must be unique')
       rules.push(
-        (v: string) => !(v.split(' ').some(r => this.excludedWordsListForSeries.includes(r.toLowerCase()))) ||
+        (v: string) => !(v?.split(' ').some(r => this.excludedWordsListForSeries.includes(r.toLowerCase()))) ||
```
severinbeauvais, line 260: Same as above (see line 251).

Is this a problem in the unit tests only, or in deployed code -- how do you reproduce this?

severinbeauvais (commented):
I'm happy with the enum changes.

I'm skeptical about the conditional chaining changes -- I don't understand why the previous code didn't work...?

Also, remember to rebuild SB.

src/components/share-structure/EditShareStructure.vue
JazzarKarim: This is a problem only in Unit tests I believe. When I ran unit tests without the optional chaining, the two tests I mentioned above were failing.

severinbeauvais: Passing unit tests are awesome, but conditional chaining needs to be used carefully as it can mask actual errors. For example, code could pass when it should actually fail.

See if you can fix the unit test (probably a data initialization problem) instead.

src/components/share-structure/EditShareStructure.vue
JazzarKarim: I agree. That's why I was confused. The code with the unit tests should work. However, as above, they don't.
without optional

severinbeauvais: Looking at the error messages, I'm guessing `s.name` is null... Maybe the data in the unit test is missing names in the shares or series objects?

severinbeauvais: Even weirder: the test failure says the error is on line 243, but your change is on line 247 !?!?

JazzarKarim: I'll be reverting the change that I made. I'll try to find some other way to fix it.

tests/unit/EditShareStructure.spec.ts
```diff
@@ -147,11 +147,7 @@ describe('Edit Share Structure component', () => {
   it('Emits add edit class event', async () => {
     const shareClass = createShareStructure(null, 1, 'Class', 'Class A', true, 100, true, 0.50, 'CAD', true)
     const wrapper: Wrapper<EditShareStructure> = createComponent(shareClass, -1, 1, null, [])
-    await Vue.nextTick()
```
JazzarKarim, line 150: Removing it (same as emitting a remove class event), fixed the tests. I also reverted the optional chaining.

severinbeauvais: That makes no sense to me. Probably it's one of those tests that doesn't check the right thing (for example, testing that something _doesn't_ exist, which passes when the code is broken).

I'm happy to ignore this for now if you want. Or if it really bugs you then I'll help you solve it, too :)

severinbeauvais: You should be able to have "await next tick" everywhere with zero impact -- it would be the same as waiting for the user to click something.

If you want to try and fix this, revert the "await next tick" and look into the emitted events, etc. I dunno.

severinbeauvais: If you don't await then maybe you're "expecting" results before the results are ready.

JazzarKarim: That makes sense. I also was surprised that removing the "await next tick" actually fixed the tests. I'm just going to revert this change (again) and merge without fixing the tests if that's OK with you. This is outside the ticket's scope I'm working on anyway 😟

severinbeauvais: LOL. Totally OK with me. Been there, done that, got the headache, took the Tylenol, lived to code another day.

severinbeauvais (approved):

tests/unit/EditShareStructure.spec.ts
JazzarKarim: Hahahaha. I FEEL YOU ON THAT SEV.

chenhongjing (approved):
LGTM 😺

PR #184, opened 2023-06-01
Make resolution optional, so we can use this for corrections.

seeker25:


seeker25:
Hold off on this one for a couple - May need to add in new payment codes for special resolution correction

JazzarKarim (approved):

PR #193, opened 2023-08-04
Add in .nojekyll file. This should allow assets with _ to be published.

seeker25:
https://github.com/storybookjs/storybook/issues/21771

" if you are publishing pages from docs folder, the .nojekyll file should be put under docs"


severinbeauvais (approved):

severinbeauvais:
I noticed that `npm run build-storybook` deletes this .nojekyll file. It must clear its output folder (docs/) when it rebuilds.

For now I'm working around this by not committing the deletion of the subject file after a Storybook rebuild.

NotJustClarkKent:
@severinbeauvais this was our issue! We added the file but the build pipeline was deleting it - thanks for the tip!!

PR #194, opened 2023-08-08
17118 update base address for vite

JazzarKarim:
*Issue #:* /bcgov/entity#17118

*Description of changes:*
- Fixed Base Address to work with Vite migration (Fixes sourcemap error in https://github.com/bcgov/business-create-ui/pull/560)
- Changed `vue-plugin-helper-decorator` to `vuelidate-property-decorators`. The former is a very outdated and old package. I've tested and the way we're using the decorator works the exact same way with both.


docs/.nojekyll
severinbeauvais, line 1: Karim, please restore this file.

severinbeauvais (commented):
I wouldn't approve all these changes on a Friday afternoon, but you're in tomorrow, right?
:)

severinbeauvais:
I thought all the tests passed previously?

docs/.nojekyll
JazzarKarim: Hmm, that's weird. It was deleted automatically when I built storybook.

JazzarKarim:
> I wouldn't approve all these changes on a Friday afternoon, but you're in tomorrow, right? :)

Sev, 99% of the changes here are from npm build storybook. The actual changes are at the end (only a handful).

severinbeauvais (commented):
Did you confirm that a sampling of these components displays correctly in Storybook (locally), with no console errors?

PS It would be nice to separate the SB files from the other changes. 179 is a large number to review (even if I skip over most of them).

severinbeauvais:
> PS It would be nice to separate the SB files from the other changes. 179 is a large number to review (even if I skip over most of them).

Actually, Travis noted recently that we should not commit the SB files, and the CI job should re-create them. We need to bug @pwei1018 to help us with this :)

package.json
```diff
@@ -64,7 +64,7 @@
     "vite-plugin-environment": "^1.1.3",
     "vitest": "^0.33.0",
     "vue-class-component": "^7.2.6",
-    "vue-plugin-helper-decorator": "^0.0.12",
+    "vuelidate-property-decorators": "1.0.28",
```
severinbeauvais, line 67: Is this needed for Storybook?

JazzarKarim:
> I thought all the tests passed previously?

Yup, they do. I re-ran the job, all is green now.

JazzarKarim:
> > PS It would be nice to separate the SB files from the other changes. 179 is a large number to review (even if I skip over most of them).
> 
> Actually, Travis noted recently that we should not commit the SB files, and the CI job should re-create them. We need to bug @pwei1018 to help us with this :)

Oh shoot. Shall I fix that? I'll only commit my changes?

JazzarKarim:
@severinbeauvais Should be much much cleaner now. No more committed SB files.

package.json
JazzarKarim: No, I don't think so! It's needed to use the `@Validations` property decorator.

docs/.nojekyll
severinbeauvais: Yeah, I think SB wipes out that folder when it rebuilds.

This is just something we need to watch our for until we fix it.

severinbeauvais: See 17345.

severinbeauvais:
> > > PS It would be nice to separate the SB files from the other changes. 179 is a large number to review (even if I skip over most of them).
> > 
> > 
> > Actually, Travis noted recently that we should not commit the SB files, and the CI job should re-create them. We need to bug @pwei1018 to help us with this :)
> 
> Oh shoot. Shall I fix that? I'll only commit my changes?

We haven't implemented the auto-Storybook-build yet. See 17345.

For now, please commit the SB files.

package.json
JazzarKarim: Actually, I take that back. It is indeed needed for Storybook. I tried running Storybook and looking at BaseAddress, it wasn't working since it failed importing the module. Sorry about that.

severinbeauvais: Interesting that SB needs it since it's actually a component dependency. Did you update the Lerna dependencies (`lerna bootstrap --hoist`)?

JazzarKarim: I did not actually. Totally forgot. You're 100% right Sev. It is not needed by SB.

src/components/base-address/package.json
```diff
@@ -11,7 +11,7 @@
     "vuelidate": "^0.7.4"
```
severinbeauvais, line 11: While you're here, can you please bump this  up to `^0.7.7`? (same as the hoisted version used by SB)

package.json
```diff
@@ -64,7 +64,6 @@
     "vite-plugin-environment": "^1.1.3",
     "vitest": "^0.33.0",
     "vue-class-component": "^7.2.6",
-    "vue-plugin-helper-decorator": "^0.0.12",
```
severinbeauvais, line 67: Darn, I tried this locally and it looks like SB need this anyway:
```
"vuelidate-property-decorators": "1.0.28",
```

JazzarKarim: Are you positive Sev? I was able to see BaseAddress on SB even when I removed this and rebuilt.

JazzarKarim: baseaddress

severinbeauvais:
I'm playing with this locally. It appears that validations are not working. I'm seeing this in the console.log when I make the address editable in SB for CompletingParty:

image

I tried adding "vuelidate-property-decorators" as a dependency for CompletingParty but that doesn't seem to help.

severinbeauvais:
> I tried adding "vuelidate-property-decorators" as a dependency for CompletingParty but that doesn't seem to help.

^^ This may or may not be needed but that's not the issue here.

The issue is with the way that validations are working. Previously, in the validation-mixin, `this.$v[prop][key].$params.maxLength` had a prop `max` with a value, but at the moment maxLength=null.

JazzarKarim:
> > I tried adding "vuelidate-property-decorators" as a dependency for CompletingParty but that doesn't seem to help.
> 
> ^^ This may or may not be needed but that's not the issue here.
> 
> The issue is with the way that validations are working. Previously, in the validation-mixin, `this.$v[prop][key].$params.maxLength` had a prop `max` with a value, but at the moment maxLength=null.

Sev, I just noticed that this issue has been present even before this PR. I went to the SB in the main repo (without my changes), and I can also see this error. I was also having this issue in my Vite PR. I fixed it by playing with the vuelidate verison. I'll check how things are going in here.

severinbeauvais:
> The issue is with the way that validations are working. Previously, in the validation-mixin, `this.$v[prop][key].$params.maxLength` had a prop `max` with a value, but at the moment maxLength=null.

We don't see this error in SB BaseAddress because no schema is declared. 

With a schema:

image

JazzarKarim:
> ^^ This may or may not be needed but that's not the issue here.
> 
> The issue is with the way that validations are working. Previously, in the validation-mixin, `this.$v[prop][key].$params.maxLength` had a prop `max` with a value, but at the moment maxLength=null.

Please try now Sev. Should be fine.

severinbeauvais:
> Please try now Sev. Should be fine.

Indeed, now it works!

Can you please add an address schema to BaseAddress? You can prob use the same one as in the CompletingParty story.

And then we'll be good to merge this!

package.json
severinbeauvais: ^^ Yeah but there's no schema, and that's what causes the validation to fail.

severinbeauvais: You fixed this with an older version of Vuelidate so this is resolved.

src/components/base-address/package.json
severinbeauvais: Obsolete comment. (It's actually an older version we need 😛 )

severinbeauvais (approved):
LGTM

severinbeauvais:
```
lerna notice
Successfully published:
 - @bcrs-shared-components/base-address@2.0.3
 - @bcrs-shared-components/completing-party@2.1.30
lerna success published 2 packages
```

JazzarKarim:
> ```
> lerna notice
> Successfully published:
>  - @bcrs-shared-components/base-address@2.0.3
>  - @bcrs-shared-components/completing-party@2.1.30
> lerna success published 2 packages
> ```

Thanks Sev!

PR #195, opened 2023-08-11
16636 Added new component "SubMenu"

severinbeauvais:
*Issue #:* bcgov/entity#16636

**I needed something like this so I coded it. In the end, it wasn't exactly what I needed, but instead of throwing away the code, I thought I'd publish it here.**

*Description of changes:*
- commit 1: added new component
- commit 2: rebuilt Storybook


severinbeauvais:
Sample:

image

src/components/sub-menu/SubMenu.vue
```diff
@@ -0,0 +1,96 @@
...
+        {{ name }}
+      </v-btn>
+    </template>
+    <v-list>
+      <template v-for="(item, index) in menuItems">
+        <v-divider
+          v-if="item.isDivider"
+          :key="index"
+        />
+        <sub-menu
```
severinbeauvais, line 43: Note the "recursion" here. This results in a submenu to the side of the current menu. What I actually wanted is a "nested" submenu, which opens in the same menu as the parent, like this:

image

src/components/sub-menu/SubMenu.vue
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Adapted from https://codepen.io/Moloth/pen/ZEBOzQP.
+ * See also https://github.com/vuetifyjs/vuetify/issues/1877.
+ */
+@Component({})
+export default class SubMenu extends Vue {
+  // props
+  @Prop() readonly name!: string
+  @Prop() readonly icon!: string
+  @Prop() readonly menuItems!: Array<any>
```
JazzarKarim, line 79: It's not possible here to specify the type of the Array right Sev? I'm looking and I'm seeing the array is of mixed things.

JazzarKarim (approved):
LGTM Sev 👍

ketaki-deodhar (approved):
LGTM!

src/components/sub-menu/SubMenu.vue
severinbeauvais: Yes, in this scenario the array is mixed items. One could come up with an interface (possibly a union of interfaces) that would work here, but I don't want to spend any more time on this as I won't be using it.

src/components/sub-menu/SubMenu.vue
severinbeauvais: I have now created a generic component that does the above ^^ and which I will reuse for Home Jurisdictions. You'll see it in a future PR :)

src/components/sub-menu/SubMenu.vue
JazzarKarim: Totally agree. I was just out asking out of curiosity. Alright Sev, resolving.

severinbeauvais:
image

PR #201, opened 2023-10-25
17123 - Upgrade Lerna to 7.4.1

ketaki-deodhar:
*Issue #:* /bcgov/entity#17123

*Description of changes:*

- Research and implement what is needed to upgrade Lerna (from ^5.6.2 to ^7.4.1). Use Lerna locally instead of globally
- Create a feature-vue3 branch. This allows us to work on these shared components separately
- update README.md
- upgrade a simple component by 1 patch version for testing

Note - Modified `DetailComment` component testing, upgrade a simple component by 1 patch version for testing and published successfully (https://www.npmjs.com/package/@bcrs-shared-components/detail-comment)
Reverted the changes made to `DetailComment` component in the last commit


README.md
```diff
@@ -73,36 +65,36 @@ Note: GitHub Pages are used to serve the Storybook app folder. As this requires
...
+./node_modules/.bin/lerna version --include-merged-tags
 ```
 
 Note: This step will create new tags and commit them!
 
-### 15. Publish to npm
+### 14. Publish to npm
 
 ```
 lerna publish from-package
```
severinbeauvais, line 92: Update this one, too.

README.md
```diff
@@ -119,7 +111,7 @@ To create a new package:
 3. add a package.json file with version = "0.0.0" (see others as examples)
 4. add the package name to the "packages" array in lerna.json
 5. commit the changes
-6. run `lerna version --include-merged-tags` to version your package to "1.0.0"
+6. run `./node_modules/.bin/lerna version --include-merged-tags` to version your package to "1.0.0"
 7. run `lerna publish from-package` to publish your package to npm
```
severinbeauvais, line 115: And this one.

README.md
```diff
@@ -34,31 +34,23 @@ git co -b my-feature-branch
 ### 4. Install dependencies
 
 ```
-npm i                         // if new repo
-npm i --global lerna@^5.6.2   // if you didn't do this previously
+npm i                                 // if new repo
+./node_modules/.bin/lerna --version   // Use Lerna locally instead of globally
```
severinbeauvais, line 38: I suggest either keeping this line and changing the comment to "verify that Lerna is installed locally" or deleting this line altogether. Thoughts?

Also it can be called out to users of this repo that Lerna doesn't need to be installed globally any more, so it can be deleted.

package.json
```diff
@@ -52,6 +52,7 @@
     "eslint-plugin-vuetify": "^1.1.0",
     "flush-promises": "^1.0.2",
     "jsdom": "^22.1.0",
+    "lerna": "^7.4.1",
```
severinbeauvais, line 55: Maybe we should add a "version" (after line 2) to this repo as well, eh?

I'm thinking either 2.0.0 since it's the first significant change to this repo's top level, or maybe 3.0.0 to coincide with upcoming Vue changes?

ketaki-deodhar: Yes makes sense. I think 3.0.0 to coincide

README.md
ketaki-deodhar: 1. change comment to avoid confusion
2. I will add a note for second

JazzarKarim (commented):
Ketaki, do we need to update the package.json versions manually after the lerna update? Can we, for example, set the version to 2.0.0 and publish?

README.md
```diff
@@ -33,32 +33,26 @@ git co -b my-feature-branch
 
 ### 4. Install dependencies
 
+**\*\*\* NOTE: Lerna doesn't need to be installed globally any more, so it can be deleted \*\*\***
```
ketaki-deodhar, line 36: Note for users and updated comment below about Lerna version (line 40)

package.json
ketaki-deodhar: Updated version to 3.0.0

ketaki-deodhar:
> Ketaki, do we need to update the package.json versions manually after the lerna update? Can we, for example, set the version to 2.0.0 and publish?

Karim, do you mean for the changed components?

JazzarKarim:
> > Ketaki, do we need to update the package.json versions manually after the lerna update? Can we, for example, set the version to 2.0.0 and publish?
> 
> Karim, do you mean for the changed components?

Yes! Let's say we upgrade DetailComponent to Vue 3 and we want to only publish that component with a major release (2.0.0). Would it be possible to do that?

ketaki-deodhar:
> > > Ketaki, do we need to update the package.json versions manually after the lerna update? Can we, for example, set the version to 2.0.0 and publish?
> > 
> > 
> > Karim, do you mean for the changed components?
> 
> Yes! Let's say we upgrade DetailComment component to Vue 3 and we want to only publish that component with a major update (2.0.0). Would it be possible to do that?

I think so but do not have to change package.json manually as far as I understand. @severinbeauvais please correct me if I got it wrong :)

severinbeauvais:
> > > > Ketaki, do we need to update the package.json versions manually after the lerna update? Can we, for example, set the version to 2.0.0 and publish?
> > > 
> > > 
> > > Karim, do you mean for the changed components?
> > 
> > 
> > Yes! Let's say we upgrade DetailComment component to Vue 3 and we want to only publish that component with a major update (2.0.0). Would it be possible to do that?
> 
> I think so but do not have to change package.json manually as far as I understand. @severinbeauvais please correct me if I got it wrong :)

The version to publish is selected at the Lerna stage. Do you remember when it prompted us for the new version? We don't touch the version in the package.json files.

JazzarKarim:
> > > > > Ketaki, do we need to update the package.json versions manually after the lerna update? Can we, for example, set the version to 2.0.0 and publish?
> > > > 
> > > > 
> > > > Karim, do you mean for the changed components?
> > > 
> > > 
> > > Yes! Let's say we upgrade DetailComment component to Vue 3 and we want to only publish that component with a major update (2.0.0). Would it be possible to do that?
> > 
> > 
> > I think so but do not have to change package.json manually as far as I understand. @severinbeauvais please correct me if I got it wrong :)
> 
> The version to publish is selected at the Lerna stage. Do you remember when it prompted us for the new version? We don't touch the version in the package.json files.

I see. So you can select which version you want, that's awesome. OK, got it.

severinbeauvais (approved):

ketaki-deodhar:
> > > > > Ketaki, do we need to update the package.json versions manually after the lerna update? Can we, for example, set the version to 2.0.0 and publish?
> > > > 
> > > > 
> > > > Karim, do you mean for the changed components?
> > > 
> > > 
> > > Yes! Let's say we upgrade DetailComment component to Vue 3 and we want to only publish that component with a major update (2.0.0). Would it be possible to do that?
> > 
> > 
> > I think so but do not have to change package.json manually as far as I understand. @severinbeauvais please correct me if I got it wrong :)
> 
> The version to publish is selected at the Lerna stage. Do you remember when it prompted us for the new version? We don't touch the version in the package.json files.

Got it!

JazzarKarim (approved):
LGTM!

PR #202, opened 2023-11-09
18513 Changed transition from "expand" to "slide-y"

severinbeauvais:
*Issue #:* bcgov/entity#18513

*Description of changes:*
Since the expand transition was jerky, I changed it to "slide-y" along with the `hide-on-leave` options.

### DIFF
--- a/src/components/expandable-help/ExpandableHelp.vue
+++ b/src/components/expandable-help/ExpandableHelp.vue
@@ -11,7 +11,7 @@
       <span class="pl-2 help-label">{{ !helpToggle ? helpLabel : 'Hide Help' }}</span>
     </div>
 
-    <v-expand-transition>
+    <v-slide-y-transition hide-on-leave>
       <div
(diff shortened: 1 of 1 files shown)


src/components/expandable-help/ExpandableHelp.vue
```diff
@@ -11,7 +11,7 @@
       <span class="pl-2 help-label">{{ !helpToggle ? helpLabel : 'Hide Help' }}</span>
     </div>
 
-    <v-expand-transition>
+    <v-slide-y-transition hide-on-leave>
```
severinbeauvais, line 14: I also tried "expand" with `hide-on-leave` (which has no effect), and "scale" (which explodes on the screen kinda weird), and "scroll-y" (which looks almost the same as slide-y. Ultimately, "slide-y" with the option looks best to me.

JazzarKarim (approved):

PR #210, opened 2023-11-27
18024 Update DatePicker to not validate while still open

leodube-aot:
*Issue #:* /bcgov/entity#18024

*Description of changes:*
- Fix UI issue described in image below
- Only evaluate `inputRules` when `dateText` is not null
- Set default for `initialValue` to be null instead of an empty string
- I think it is ok if error message remains after initial input attempt (ie. They open the datepicker, click "Ok" or "Cancel" for an invalid date, then when they re-open the picker the validation message will not get cleared).
- New story: `datePickerValidation` can be used for testing the fix

285523902-9ff92583-80f6-4786-8804-433ffc2c7d4c


JazzarKarim (approved):
LGTM! 👍

severinbeauvais (approved):

severinbeauvais:
```
lerna notice
Successfully published:
 - @bcrs-shared-components/approval-type@1.0.42
 - @bcrs-shared-components/date-picker@1.2.39
 - @bcrs-shared-components/effective-date-time@1.1.39
lerna success published 3 packages
```

PR #212, opened 2023-11-30
18526 - import Date picker components

ketaki-deodhar:
*Issue #:* /bcgov/entity#18526

*Description of changes:*

- Create new components RangeDatePicker, InputFieldDatePicker, BaseDatePicker
- update libraries to make unit tests work

BaseDatePicker: 

image

InputFieldDatePicker: 

https://github.com/bcgov/bcrs-shared-components/assets/116035339/4a3e7e66-fe45-4164-993e-baa38774ce04

RangeDatePicker:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
     "@storybook/addon-essentials": "^7.5.3",
     "@storybook/vue3": "^7.5.3",
     "@storybook/vue3-vite": "^7.5.3",
-    "@testing-library/jest-dom": "6.1.4",
     "@types/lodash": "^4.14.194",
     "@types/sinon": "^10.0.14",
     "@types/vue-the-mask": "^0.11.1",
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "@bcrs-shared-components/mixins": "^1.1.16",
     "@mdi/font": "^5.9.55",
+    "@vuepic/vue-datepicker": "^7.2.2",
     "axios": "^0.27.2",
     "country-list": "^2.3.0",
     "lodash": "4.17.21",
@@ -25,14 +26,14 @@
     "vue-router": "^3.6.5",
     "vue-the-mask": "^0.11.1",
     "vuelidate": "^0.7.4",
-    "vuetify": "^3.1.8",
-    "@vuepic/vue-datepicker": "^7.2.2"
+    "vuetify": "^3.1.8"
   },
   "devDependencies": {
     "@mdx-js/react": "^2.3.0",
     "@storybook/addon-essentials": "^7.5.3",
     "@storybook/vue3": "^7.5.3",
     "@storybook/vue3-vite": "^7.5.3",
+    "@testing-library/jest-dom": "6.1.4",
     "@types/lodash": "^4.14.194",
     "@types/sinon": "^10.0.14",
     "@types/vue-the-mask": "^0.11.1",
@@ -43,7 +44,7 @@
     "@volar-plugins/vetur": "latest",
     "@vue/eslint-config-standard": "^4.0.0",
     "@vue/eslint-config-typescript": "^9.1.0",
-    "@vue/test-utils": "^1.3.5",
+    "@vue/test-utils": "^2.4.0-alpha.2",
     "axios-mock-adapter": "^1.21.5",
     "eslint": "~8.50.0",
     "eslint-plugin-import": "^2.28.1",
@@ -65,9 +66,9 @@
     "vite-plugin-environment": "^1.1.3",
     "vite-plugin-vuetify": "^1.0.2",
     "vitest": "^0.33.0",
+    "vitest-axe": "^0.1.0",
     "vue-facing-decorator": "^3.0.3",
     "vue-style-loader": "^4.1.3",
-    "vue-template-compiler": "^2.7.14",
     "vue-test-utils-helpers": "git+https://github.com/bcgov/vue-test-utils-helpers.git"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bcrs-shared-components",
-  "version": "3.0.0",
+  "version": "3.0.1",
   "private": true,
   "appName": "BCRS Shared Components",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "@vue/test-utils": "^2.4.0-alpha.2",
     "axios-mock-adapter": "^1.21.5",
     "eslint": "~8.50.0",
-    "eslint-plugin-import": "^2.28.1",
     "eslint-plugin-storybook": "^0.6.15",
     "eslint-plugin-vue": "^9.17.0",
     "eslint-plugin-vuetify": "^2.0.2",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "vue-router": "^3.6.5",
     "vue-the-mask": "^0.11.1",
     "vuelidate": "^0.7.4",
-    "vuetify": "^3.1.8"
+    "vuetify": "^3.1.8",
+    "@vuepic/vue-datepicker": "^7.2.2"
   },
   "devDependencies": {
     "@mdx-js/react": "^2.3.0",
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "@vue/test-utils": "^1.3.5",
     "axios-mock-adapter": "^1.21.5",
     "eslint": "~8.50.0",
+    "eslint-plugin-import": "^2.28.1",
     "eslint-plugin-storybook": "^0.6.15",
     "eslint-plugin-vue": "^9.17.0",
     "eslint-plugin-vuetify": "^2.0.2",
--- a/src/components/base-date-picker/BaseDatePicker.vue
+++ b/src/components/base-date-picker/BaseDatePicker.vue
@@ -1,0 +1,407 @@
+<template>
+  <div
+    class="base-date-picker"
+    :class="{ 'base-date-picker__err': error }"
+  >
+    <div class="base-date-picker__header">
+      <v-btn
+        class="base-date-picker__header__year"
+        variant="text"
+        @click="openYearsSelection = true"
+      >
+        {{ selectedYear }}
+      </v-btn>
+      <div
+        v-if="selectedDate"
+        class="base-date-picker__header__date"
+      >
+        {{ getWeekDay(selectedDate) }}, {{ getMonth(selectedDate) }} {{ selectedDate.getDate() }}
+      </div>
+    </div>
+    <VueDatePicker
+      v-model="selectedDate"
+      autoApply
+      :actionRow="{ showCancel: false, showNow: false, showPreview: false, showSelect: false }"
+      calendarCellClassName="base-date-picker__calendar__day"
+      calendarClassName="base-date-picker__calendar"
+      :dayNames="['SUN', 'MON','TUE','WED','THU','FRI','SAT']"
+      :enableTimePicker="enableTimePicker"
+      format="yyyy-MM-dd"
+      hideOffsetDates
+      inline
+      :maxDate="maxDate"
+      :minDate="minDate"
+      :monthChangeOnScroll="false"
+      :weekStart="0"
+    >
+      <template
+        #month-year="{
+          month,
+          year,
+          months,
+          years,
+          handleMonthYearChange,
+          updateMonthYear
+        } : {
+          month: number,
+          year: number,
+          months: { text: string, value: number }[],
+          years: { text: string, value: number }[],
+          handleMonthYearChange: (isNext: boolean) => void,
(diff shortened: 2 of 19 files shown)


tests/unit/DetailComment.spec.ts
```diff
@@ -42,49 +39,40 @@ describe('DetailComment', () => {
...
-    vm.emitInput('testing 4 5 6')
+    const element = wrapper.find('#detail-comment-textarea')
+    element.setValue('testing 4 5 6')
 
     // verify valid and input events
     expect(wrapper.emitted('valid').pop()[0]).toEqual(true)
-    expect(wrapper.emitted('input').pop()).toEqual(['testing 4 5 6'])
-
-    wrapper.destroy()
+    // expect(wrapper.emitted('updateValue')[0]).toEqual(['testing 4 5 6'])
```
ketaki-deodhar, line 76: Couldn't get this working but will keep looking for solution.......

package.json
```diff
@@ -42,9 +43,10 @@
     "@volar-plugins/vetur": "latest",
     "@vue/eslint-config-standard": "^4.0.0",
     "@vue/eslint-config-typescript": "^9.1.0",
-    "@vue/test-utils": "^1.3.5",
+    "@vue/test-utils": "^2.4.0-alpha.2",
```
ketaki-deodhar, line 46: needed to install `@vue/test-utils@next` to get compatible version of `vue-test-utils` with vue 3

src/components/base-date-picker/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@bcrs-shared-components/base-date-picker",
+  "version": "0.0.0",
```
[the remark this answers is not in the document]
severinbeauvais: No, leave as 0.0.0 for your PR. When I publish this, it will get set to 1.0.0.

package.json
```diff
@@ -42,9 +43,10 @@
...
     "axios-mock-adapter": "^1.21.5",
     "eslint": "~8.50.0",
+    "eslint-plugin-import": "^2.28.1",
```
severinbeauvais, line 49: What's this for?

package.json
```diff
@@ -63,9 +65,9 @@
     "vite-plugin-environment": "^1.1.3",
     "vite-plugin-vuetify": "^1.0.2",
     "vitest": "^0.33.0",
+    "vitest-axe": "^0.1.0",
```
severinbeauvais, line 68: What's this for?

src/components/base-date-picker/BaseDatePicker.vue
severinbeauvais, line 1: Did you change anything in here or is it the same as what we got from Cameron?

src/components/input-field-date-picker/InputFieldDatePicker.vue
severinbeauvais, line 1: This looks similar to our original date picker. Does it fully replace it?

src/components/range-date-picker/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "@bcrs-shared-components/range-date-picker",
+  "version": "0.0.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "lodash": "4.17.21"
+  },
+  "devDependencies": {
+    "vue-facing-decorator": "^3.0.3"
```
severinbeauvais, line 11: Looks like it doesn't depend on  lodash or vue-facing-decorator.

Looks like it depends on:
- vue
- shared base-date-picker

tests/unit/DetailComment.spec.ts
```diff
@@ -42,49 +39,40 @@ describe('DetailComment', () => {
     expect(vm.value).toBe('Initial comment')
     expect(vm.placeholder).toBe('Enter Comment Here')
     expect(vm.autofocus).toBe(true)
-
-    wrapper.destroy()
   })
 
   // FUTURE: fix this test
```
severinbeauvais, line 44: Does this test need more fixing or is it good now?

tests/unit/DetailComment.spec.ts
severinbeauvais: You may have to trigger an update on the textarea (submit or blur maybe?), to get it to recognize the new value you set. Do a global search (maybe in other UIs, too) for `setValue` and see what was done.

package.json
ketaki-deodhar: This is to import vitest-axe/extend-expect in test setup file to make sure tests work with Vitest

Ref: https://github.com/chaance/vitest-axe

package.json
ketaki-deodhar: Something I added for resolving an issue with imports but I think this is not useful and will remove it.

src/components/base-date-picker/BaseDatePicker.vue
ketaki-deodhar: Pretty much the same. Only change was to use correct css for validation message.

src/components/input-field-date-picker/InputFieldDatePicker.vue
ketaki-deodhar: Almost same. There are slight differences.
Main branch/DEV:

image

Label
image

Vue3 branch:

Style is a bit different
image

Does not have label
image

src/components/input-field-date-picker/InputFieldDatePicker.vue
```diff
@@ -0,0 +1,174 @@
...
+      emitCancel,
+      validate,
+      isDateValid,
+      ...toRefs(localState)
+    }
+  }
+})
+</script>
+
+<style lang="scss" scoped>
```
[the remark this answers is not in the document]
ketaki-deodhar: This is in DEV/Main branch:
image

The wrapper was wider and the base date picker width was smaller. This is fixed now to look same as above.
image

Do you think it should be smaller?

ketaki-deodhar:
> Check with Cam on whether he has any date component unit tests you could add here.

I checked with Cameron and unfortunately they do not have any unit tests. Does it make sense to create a new ticket for building unit tests?

src/components/input-field-date-picker/InputFieldDatePicker.vue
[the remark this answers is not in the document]
severinbeauvais: UI/UX will have final say on size. I was commenting on this (below) but you've fixed it now :)

image

src/components/input-field-date-picker/InputFieldDatePicker.vue
severinbeauvais: @yuisotozaki Please see differences above between our Vue2 date picker component and the Vue3 one. (Underneath, they're completing different code.) Should we spend any time making the Vue3 one look the same or can we live with the new one as-is?

yuisotozaki: @severinbeauvais The plan is to try to not customize as much as we can. We'll want to unify the primary colour of the buttons in Vue3 date picker popup. As for the label that tuck away, I'll check with the design team today. It's a bad timing right now as we're waiting on GDX design system release and Figma licensing. FYI: Alexis Ronse is our point person for all design system things going forward.

severinbeauvais: What I'm hearing is that we won't change the Vue3 data picker above right now.

Note that we have (or will shortly have) two date pickers to choose from: the Vuetify date picker wrapper (same thing we did in Vue2) and the custom-built one from Assets team (as above). Ultimately we should winnow that down to 1 but that's a future concern.

leodube-aot: Just to add to Sev's comment, my current ticket (bcgov/entity#18779) is for updating our existing Vue 2 DatePicker component to Vue 3, so probably no need to try to get BaseDatePicker to look like our existing one.

severinbeauvais: Leo, it could go either way: either we make yours look like BaseDatePicker, or we make BaseDatePicker looks like yours, or we leave them different.

One reason for leaving BaseDatePicker alone is to keep the code as we got it. This would allow Assets team to use this shared component with no changes.

Maybe the question for @yuisotozaki is, "which date picker should we use when we update our UIs to Vue3 -- the Vuetify one or the custom one?"

severinbeauvais: Personally, I'd prefer to use the Vuetify one (assuming it's compatible), as it's less code for us to manage. Though it is more risky, and we do have the source code for the custom one...

tests/unit/DetailComment.spec.ts
ketaki-deodhar: I haven't been able to resolve this and looking into it. I would like to merge this PR first so that other developers can start updating unit tests for the components they are working on. I will keep a note of this and work on this unit test later .

leodube-aot: @ketaki-deodhar Not sure if you've found this already, but this comment from filings UI might help. Apparently you can't set a textarea's value so you might have to do it explicitly: https://github.com/bcgov/business-filings-ui/blob/main/tests/unit/DetailComment.spec.ts#L75-L82

leodube-aot (approved):
LGTM

tests/unit/DetailComment.spec.ts
severinbeauvais: Sounds good to me.

Don't spend too long (don't get stuck) on anything. For now we just need to make things like-for-like in Vue3. Thanks.

tests/unit/DetailComment.spec.ts
```diff
@@ -42,49 +42,38 @@ describe('DetailComment', () => {
...
   })
 
-  // FUTURE: fix this test
-  it.skip('emits valid event when value prop is changed', async () => {
-    const wrapper = shallowMount(DetailComment,
+  it('emits valid event when value prop is changed', async () => {
+    const wrapper = mount(DetailComment,
       {
+        props: {
+          value: 'testing 1 2 3'
```
severinbeauvais, line 51: This isn't the same as the previous test.

Previously, we started with no props and then set them. Now, you start with initial props.

I'd prefer to see the previous unit test left alone, skipped if you can't get it working. Add a new test if needed.

Also, why did you remove old line 65?

tests/unit/DetailComment.spec.ts
```diff
@@ -42,49 +42,38 @@ describe('DetailComment', () => {
...
-
-    wrapper.destroy()
   })
 
   it('emits events when value model is changed', async () => {
-    const wrapper = shallowMount(DetailComment,
+    const wrapper = mount(DetailComment,
       {
+        props: {
+          value: 'testing 1 2 3'
```
severinbeauvais, line 67: Same comment as above.

src/components/base-date-picker/package.json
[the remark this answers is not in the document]
severinbeauvais: Fixed, resolving.

severinbeauvais (approved):
Figure out what you're going to do with the tests and then this is good to merge.

tests/unit/DetailComment.spec.ts
```diff
@@ -84,7 +80,5 @@ describe('DetailComment', () => {
     // verify valid and input events
     expect(wrapper.emitted('valid').pop()[0]).toEqual(true)
     expect(wrapper.emitted('input').pop()).toEqual(['testing 4 5 6'])
-
-    wrapper.destroy()
```
severinbeauvais, line 88: Is there a good reason for removing this? It was initially added to prevent a memory leak, but maybe is no longer needed.

tests/unit/DetailComment.spec.ts
ketaki-deodhar: For this unit test, earlier there was a line `wrapper.setProps({ value: 'testing 1 2 3' })` which I just changed to setting prop while mounting.  I will revert it back and skip the test.

tests/unit/DetailComment.spec.ts
ketaki-deodhar: Actually destroy() is now changed to unmount(). Changed it accordingly.
Ref: https://test-utils.vuejs.org/migration/#destroy-is-now-unmount-to-match-Vue-3

tests/unit/DetailComment.spec.ts
severinbeauvais: Sure.

"setProps" is different from mounting the component with props. In the former, you are _changing_ the props. In the latter, they are initialized with those values. The former can cause getters and watchers to react.

severinbeauvais: Thanks for the new test (in addition to the original, skipped test).

FWIW, I consider this conversation resolved.

tests/unit/DetailComment.spec.ts
severinbeauvais: 🤯 

I guess this will have to be changed everywhere. @leodube-aot @PaulGarewal @JazzarKarim, FYI.

PR #214, opened 2023-12-05
18752 base-address upgrade

PaulGarewal:
*Issue #:* /bcgov/entity#18752

*Description of changes:*

- Added new vue3 base address from Assets Team. 
- Testing still needed to be completed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "bcrs-shared-components",
-  "version": "3.0.1",
+  "version": "3.0.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "bcrs-shared-components",
-      "version": "3.0.1",
+      "version": "3.0.3",
       "dependencies": {
         "@bcrs-shared-components/mixins": "^1.1.16",
         "@mdi/font": "^5.9.55",
@@ -22,7 +22,7 @@
         "vue-router": "^3.6.5",
         "vue-the-mask": "^0.11.1",
         "vuelidate": "^0.7.4",
-        "vuetify": "^3.1.8"
+        "vuetify": "^3.4.5"
       },
       "devDependencies": {
         "@mdx-js/react": "^2.3.0",
@@ -21241,9 +21241,9 @@
       }
     },
     "node_modules/vuetify": {
-      "version": "3.3.23",
-      "resolved": "https://registry.npmjs.org/vuetify/-/vuetify-3.3.23.tgz",
-      "integrity": "sha512-EVEFvDyKr/HK2+hQJyhQSjIDULDaBi9vfgAAB0ekm+2Tvc8AkaDwb7VK44R2eilGFGpju6iN52FV7Uasj8SmIg==",
+      "version": "3.4.6",
+      "resolved": "https://registry.npmjs.org/vuetify/-/vuetify-3.4.6.tgz",
+      "integrity": "sha512-Zp6BjKkNcnyHlOmr01Y/Rc4d686PORO+hcpfM76qRaDMIU4H1GnP0JIF3jwxI1q3Uj/2bHKLwn81oCLJZd+8sg==",
       "engines": {
         "node": "^12.20 || >=14.13"
       },
@@ -21254,7 +21254,7 @@
       "peerDependencies": {
         "typescript": ">=4.7",
         "vite-plugin-vuetify": "^1.0.0-alpha.12",
-        "vue": "^3.2.0",
+        "vue": "^3.3.0",
         "vue-i18n": "^9.0.0",
         "webpack-plugin-vuetify": "^2.0.0-alpha.11"
       },
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,6 @@
 {
   "name": "bcrs-shared-components",
   "version": "3.0.1",
-  "version": "3.0.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
@@ -7255,48 +7254,6 @@
       },
       "peerDependencies": {
         "vue": ">=3.2.0"
-      }
-    },
-    "node_modules/@vuelidate/core": {
-      "version": "2.0.3",
-      "resolved": "https://registry.npmjs.org/@vuelidate/core/-/core-2.0.3.tgz",
-      "integrity": "sha512-AN6l7KF7+mEfyWG0doT96z+47ljwPpZfi9/JrNMkOGLFv27XVZvKzRLXlmDPQjPl/wOB1GNnHuc54jlCLRNqGA==",
-      "dependencies": {
-        "vue-demi": "^0.13.11"
-      },
-      "peerDependencies": {
-        "@vue/composition-api": "^1.0.0-rc.1",
-        "vue": "^2.0.0 || >=3.0.0"
-      },
-      "peerDependenciesMeta": {
-        "@vue/composition-api": {
-          "optional": true
-        }
-      }
-    },
-    "node_modules/@vuelidate/core/node_modules/vue-demi": {
-      "version": "0.13.11",
-      "resolved": "https://registry.npmjs.org/vue-demi/-/vue-demi-0.13.11.tgz",
-      "integrity": "sha512-IR8HoEEGM65YY3ZJYAjMlKygDQn25D5ajNFNoKh9RSDMQtlzCxtfQjdQgv9jjK+m3377SsJXY8ysq8kLCZL25A==",
-      "hasInstallScript": true,
-      "bin": {
-        "vue-demi-fix": "bin/vue-demi-fix.js",
-        "vue-demi-switch": "bin/vue-demi-switch.js"
-      },
-      "engines": {
-        "node": ">=12"
-      },
-      "funding": {
-        "url": "https://github.com/sponsors/antfu"
-      },
-      "peerDependencies": {
-        "@vue/composition-api": "^1.0.0-rc.1",
-        "vue": "^3.0.0-0 || ^2.6.0"
-      },
-      "peerDependenciesMeta": {
-        "@vue/composition-api": {
-          "optional": true
-        }
       }
     },
     "node_modules/@vuetify/loader-shared": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,7 @@
 {
   "name": "bcrs-shared-components",
   "version": "3.0.1",
+  "version": "3.0.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
@@ -10,7 +11,7 @@
       "dependencies": {
         "@bcrs-shared-components/mixins": "^1.1.16",
         "@mdi/font": "^5.9.55",
-        "@vuelidate/core": "^2.0.3",
+        "@vuepic/vue-datepicker": "^7.2.2",
         "axios": "^0.27.2",
         "country-list": "^2.3.0",
         "lodash": "4.17.21",
@@ -39,7 +40,7 @@
         "@volar-plugins/vetur": "latest",
         "@vue/eslint-config-standard": "^4.0.0",
         "@vue/eslint-config-typescript": "^9.1.0",
-        "@vue/test-utils": "^1.3.5",
+        "@vue/test-utils": "^2.4.0-alpha.2",
         "axios-mock-adapter": "^1.21.5",
         "eslint": "~8.50.0",
         "eslint-plugin-storybook": "^0.6.15",
@@ -60,9 +61,9 @@
         "vite-plugin-environment": "^1.1.3",
         "vite-plugin-vuetify": "^1.0.2",
         "vitest": "^0.33.0",
+        "vitest-axe": "^0.1.0",
         "vue-facing-decorator": "^3.0.3",
         "vue-style-loader": "^4.1.3",
-        "vue-template-compiler": "^2.7.14",
         "vue-test-utils-helpers": "git+https://github.com/bcgov/vue-test-utils-helpers.git"
       }
     },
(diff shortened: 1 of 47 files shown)


PaulGarewal:
Not sure if this is what we are looking for:

image

src/components/base-address/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bcrs-shared-components/base-address",
-  "version": "2.0.9",
+  "version": "2.0.10",
```
severinbeauvais, line 3: Don't update this -- the publish step (done after your PR is merged) handles this.

Also, we'll want to change the major version number!

src/components/base-address/interfaces/address-interface.ts
```diff
@@ -0,0 +1,9 @@
+export interface AddressIF {
+  street: string
+  streetAdditional?: string
+  city: string
+  region: string
+  postalCode: string
+  country: string
+  deliveryInstructions?: string
+}
```
severinbeauvais, line 9: Is this an internal-only interface? If not then it should be in the interfaces folder so that UIs can import it. (There may be an address interface there already. Can it be reused?)

severinbeauvais:
You will know you have succeeded when this component can be used externally and works properly there. Hopefully it can be used in the same way as the old component. You may need to create a very simple Vue3 UI to test this.

JazzarKarim (approved):
LGTM as long as you've tested it Paul 👍

severinbeauvais:
> Not sure if this is what we are looking for (but it is a start):

So there's no labels on the fields? Ask UI if we can use it like this (which is presumably the same as in the Assets/PPR UI).

PaulGarewal:
> > Not sure if this is what we are looking for (but it is a start):
> 
> So there's no labels on the fields? Ask UI if we can use it like this (which is presumably the same as in the Assets/PPR UI).

Ok will do. Here is what the fields look like before the validation:

image

PaulGarewal:
> > > Not sure if this is what we are looking for (but it is a start):
> > 
> > 
> > So there's no labels on the fields? Ask UI if we can use it like this (which is presumably the same as in the Assets/PPR UI).
> 
> Ok will do. I will ask Yui.
> 
> Here is what the fields look like before entering any values (for reference):
> 
> image

Confirmed with Yui that we want the same setup as the Assets Team for the BaseAddress.

severinbeauvais:
> Confirmed with Yui that we want the same setup as the Assets Team for the BaseAddress.

Nice! I hoped this would be the case but now it's official.

severinbeauvais:
> You will know you have succeeded when this component can be used externally and works properly there. Hopefully it can be used in the same way as the old component. You may need to create a very simple Vue3 UI to test this.

Any change this will work in a Vue2 UI?

Can you do an address lookup in Storybook? (I know it needs an API key, and we won't want to put that key in any commit in GH as our repos are public!)

Are the props and events the same as the old address component?

cameron-eyds:
@severinbeauvais @PaulGarewal 
If your keen on testing in a Vue3 instance before committing, feel free to pull our feature branch:
https://github.com/bcgov/ppr/tree/Vue3-Upgrade-Feature-Branch

It's pretty stable, for the most part :) 
Hit me up for an .env if you want to run it. 
OR if you want to commit and give it a go as a true import, the above holds true as well 👍

package.json
```diff
@@ -14,6 +14,7 @@
   "dependencies": {
     "@bcrs-shared-components/mixins": "^1.1.16",
     "@mdi/font": "^5.9.55",
+    "@vuelidate/core": "^2.0.3",
```
severinbeauvais, line 17: Since you're changing this file, please update the patch version on line 3.

src/components/base-address/BaseAddress.vue
severinbeauvais, line 1: I think you're missing "base-address/index.ts".

src/components/base-address/enums/validation-rule.ts
severinbeauvais, line 1: Is this internal only, or will an app need to import this?

If so then you have 2 options:
1. export everything needed in this "base-address/index.ts"
2. move this to shared enums (maybe with a more specific name like address-validation-rules.ts)

We haven't used option 1 yet in this repo but it feels like a good idea to keep all related things together. What do you think?

src/components/base-address/factories/countries-provinces-factory.ts
```diff
@@ -0,0 +1,78 @@
+import countriesData from 'country-list/data.json'
+import provincesData from 'provinces/provinces.json'
```
severinbeauvais, line 2: You need a "base-address/package.json" (see other components for examples; set version to 0.0.0 initially) to specify all this component's dependencies.

Check the other files in here for additional dependencies.

package.json
severinbeauvais: Wait - this file is used by Storybook. Does SB need this package, or is it needed by the component only?

src/components/base-address/BaseAddress.vue
PaulGarewal: It is there:

image

src/interfaces/addresses-interface.ts
```diff
@@ -4,6 +4,7 @@
  * https://github.com/bcgov/business-schemas/blob/main/src/registry_schemas/schemas/address.json
  */
 export interface AddressIF {
+[x: string]: any
```
severinbeauvais, line 7: Move this last, and indent it.

(It's a cheat to allow undeclared properties in this interface/type. Undesirable but we're not going to fix this right now.)

src/components/base-address/BaseAddress.vue
severinbeauvais: Oops, missed it. Sorry and thanks!

severinbeauvais: If an app wants to use this component, and it needs some enums/interfaces/etc that we decided to keep in this folder, then those should be export above as well.

src/components/base-address/enums/validation-rule.ts
PaulGarewal: Option 2 it is 👍

src/components/base-address/factories/countries-provinces-factory.ts
PaulGarewal: This is the current base-address/package.json:
image
Should I reset the version to 0.0.0 and add the relevant dependencies?

src/components/base-address/factories/validation-factory.ts
```diff
@@ -1,6 +1,6 @@
 import { ref } from 'vue-demi'
 
-import { ValidationRule } from '@/components/base-address/enums'
+import { ValidationRule } from '@/enums/address-validation-rules'
```
severinbeauvais, line 3: This should be `@bcrs-shared-components/enums` otherwise here (ie, when Storybook uses this component), it resolves to the local enums folder, but when used in an app, it will resolve to the app's enums folder.

severinbeauvais: This is a problem here: https://github.com/bcgov/bcrs-shared-components/blob/4c62c41230422c5ceef744176fdc1bb04d6b5372/src/components/date-picker/DatePicker.vue#L84

But Leo said he got it working: https://github.com/bcgov/bcrs-shared-components/pull/215#discussion_r1417854456

src/enums/address-validation-rules.ts
```diff
@@ -0,0 +1,9 @@
+export enum ValidationRule {
```
severinbeauvais, line 1: Ideally the enum name should match the filename.

Also, our convention is that enums are plural.

src/components/base-address/factories/countries-provinces-factory.ts
severinbeauvais: Oh right, you're updating an existing component.

Leave the version alone and I'll bump it to 3.0.0 when I publish this.

package.json
severinbeauvais: Yes, scope this import to the component itself.

src/components/base-address/interfaces/address-interface.ts
severinbeauvais: Fixed, thanks!

src/components/base-address/factories/validation-factory.ts
```diff
@@ -1,6 +1,6 @@
 import { ref } from 'vue-demi'
 
-import { ValidationRule } from '@/enums/address-validation-rules'
+import { ValidationRule } from '@bcrs-shared-components/enums/address-validation-rules'
```
severinbeauvais: If you've add address-validation-rules to enums/index.ts then you can just import the "folder". See how other enums/interfaces are exported/imported.

src/components/base-address/BaseAddress.vue
severinbeauvais, line 1: I think you need to update line 172 also.

src/components/base-address/BaseAddress.vue
```diff
@@ -243,8 +243,8 @@ export default defineComponent({
         localSchema.postalCode = origPostalCodeRules.concat([baseRules.zipCode])
         localSchema.region = origRegionRules
       } else {
-        localSchema.postalCode = origPostalCodeRules.concat([baseRules.maxLength(15)])
-        localSchema.region = [baseRules.maxLength(2), ...spaceRules]
+        localSchema.postalCode = origPostalCodeRules.concat([baseRules[AddressValidationRules.MAX_LENGTH](15)])
```
severinbeauvais, line 246: Do you need to import AddressValidationRules ?

src/components/base-address/factories/validation-factory.ts
severinbeauvais: Fixed, thanks!

src/components/base-address/factories/validation-factory.ts
severinbeauvais: Beautiful, thanks!

What about the 2 interfaces?

src/components/base-address/BaseAddress.vue
PaulGarewal: It's not valid anymore, attempting to fix

src/components/base-address/BaseAddress.vue
severinbeauvais: Again, `@/` resolves to the local interfaces folder, which will work for Storybook (the local UI, sort of) but won't work when this component is imported into an app (unless the app's local interfaces folder also has these interfaces but that's undesirable).

I think these 2 interfaces should be in the shared interfaces folder and exported along with the other (unless they're absolutely only ever used internally).

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bcrs-shared-components",
-  "version": "3.0.0",
+  "version": "3.0.1",
```
severinbeauvais, line 3: Since you're no longer changing this file, please revert the version change 😛

package-lock.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "bcrs-shared-components",
   "version": "3.0.1",
+  "version": "3.0.1",
```
severinbeauvais, line 4: Merge issue. Delete this duplicate and run `npm i` again to fix it.

src/components/base-address/factories/validation-factory.ts
severinbeauvais: @PaulGarewal Should the 2 interfaces be in the shared folder?

src/components/base-address/BaseAddress.vue
severinbeauvais: Is this fixed?

src/components/base-address/BaseAddress.vue
severinbeauvais: I was thinking that the imports on lines 172 (and 173) would be from the shared versions.

src/components/base-address/BaseAddress.stories.ts
```diff
@@ -1,6 +1,6 @@
 import { Meta } from '@storybook/vue3'
 import BaseAddress from './BaseAddress.vue'
-import { AddressIF } from './interfaces/address-interface'
+import { AddressIF } from '@/interfaces'
```
severinbeauvais, line 3: Line 3 should be from shared component (I think).

src/components/base-address/BaseAddress.vue
severinbeauvais: Also elsewhere in the vue and ts files in here.

src/components/base-address/resources/default-schema.ts
```diff
@@ -1,5 +1,5 @@
-import { AddressValidationRules } from '@bcrs-shared-components/enums'
-import { SchemaIF } from '@/components/base-address/interfaces'
+import { AddressValidationRules } from '@/enums'
+import { SchemaIF } from '@/interfaces'
 import { baseRules, spaceRules } from '@/components/base-address/factories/validation-factory'
```
severinbeauvais, line 3: I'm thinking this import should be local, as in `from "./factories/validation-factory"`. This is because the factories are internal and local to this component.

The reason I'm going on about this is that [`@/` maps to the source folder in THIS repo](https://github.com/bcgov/bcrs-shared-components/blob/4c62c41230422c5ceef744176fdc1bb04d6b5372/vite.config.ts#L15), but when these components are imported (as source files) in a UI, the UI's source folder won't have the same files. Therefore, these components should import from the "shared version" of enums or interfaces or whatever they need. Then these components should work correctly in this repo (ie, in Storybook and unit tests) and also in whatever UI they're used in.

cc: @leodube-aot

src/components/base-address/BaseAddress.vue
severinbeauvais: What error?
Your component should work in Storybook, both locally and after merge.

PaulGarewal: image

src/components/base-address/resources/default-schema.ts
leodube-aot: This import should probably be `from '../factories/validation-factory'`. That resolved the error you are referencing Paul when I tested out your PR locally.

src/components/base-address/BaseAddress.vue
leodube-aot: See my comment below about fixing the`validation-factory` import. Once that is fixed you should be okay to ignore these import errors, as the component still runs in storybook fine for me. It looks to me like the IDE can't resolve the import, but is able to at compile/run.

src/components/base-address/resources/default-schema.ts
PaulGarewal: Resolved the error and I ignored the other import errors. Storybook still runs now without any problems 👍 Thank you Leo.

severinbeauvais (approved):
LGTM!

If this component works correctly in Storybook locally (except for lookup, as previously discussed) then go ahead and merge this, and then I'll publish it.

leodube-aot (approved):
Good work on this Paul!

PR #215, opened 2023-12-06
18779 Use Vuetify 3 DatePicker

leodube-aot:
*Issue #:* /bcgov/entity#18779

*Description of changes:*
- Updated Vuetify 3 to version with DatePicker component
- Updated our DatePicker component to use the new one from Vuetify 3. The storybook stories seem to all work the same as our previous DatePicker
- Updated storybook files
- I'll rebase and ~finish up unit tests~ _(tech debt ticket created instead)_ once Ketaki's PR (#212 ) is merged as she has fixed up the unit test setup

### DIFF
--- a/src/components/date-picker/DatePicker.vue
+++ b/src/components/date-picker/DatePicker.vue
@@ -75,7 +75,7 @@
 </template>
 
 <script lang="ts">
-import { Component, Emit, Prop, Watch } from 'vue-facing-decorator'
+import { Component, Emit, Prop, Watch, Vue } from 'vue-facing-decorator'
 import { FormIF } from '@bcrs-shared-components/interfaces'
 import { DateMixin } from '@bcrs-shared-components/mixins'
 
@@ -127,9 +127,25 @@ export default class DatePicker extends DateMixin {
     this.date = this.yyyyMmDdToDate(this.initialValue)
   }
 
+  /** The formatted picker date string (YYYY-MM-DD) */
+  get pickerDate (): string {
+    if (!this.date) return ''
+
+    const dateStr = this.date.toLocaleDateString('en-US', {
+      month: 'numeric', // 12
+      day: 'numeric', // 31
+      year: 'numeric' // 2020
+    })
+
+    // convert mm/dd/yyyy to yyyy-mm-dd
+    // and make sure month and day are 2 digits (eg, 03)
+    const [ mm, dd, yyyy ] = dateStr.split('/')
+    return `${yyyy}-${mm.padStart(2, '0')}-${dd.padStart(2, '0')}`
+  }
+
   /** The display Date. */
   get displayDate (): string {
-    return this.dateToPacificDate(this.date, true)
+    return this.yyyyMmDdToPacificDate(this.pickerDate, true)
   }
 
   /** True when the picker is not displayed or disabled. */
@@ -141,7 +157,7 @@ export default class DatePicker extends DateMixin {
   @Emit('emitDate')
   protected emitDate (): string {
     this.displayPicker = false
-    return this.dateToYyyyMmDd(this.date)
+    return this.pickerDate
   }
 
   /** Emit cancel event and clear the date. */
@@ -160,7 +176,7 @@ export default class DatePicker extends DateMixin {
   @Emit('emitDateSync')
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   private emitDateSync (): string {
-    return this.dateToYyyyMmDd(this.date)
+    return this.pickerDate
   }
 
   @Watch('$route')
--- a/src/components/date-picker/DatePicker.vue
+++ b/src/components/date-picker/DatePicker.vue
@@ -133,7 +133,7 @@ export default class DatePicker extends DateMixin {
 
   /** The display Date. */
   get displayDate (): string {
-    return this.dateToPacificDate(this.date, true)
+    return this.dateToPacificDate(this.date, true, false, false)
   }
 
   /** True when the picker is not displayed or disabled. */
--- a/src/components/date-picker/DatePicker.vue
+++ b/src/components/date-picker/DatePicker.vue
@@ -127,7 +127,12 @@ export default class DatePicker extends DateMixin {
     this.date = this.yyyyMmDdToDate(this.initialValue)
   }
 
-  /** The formatted picker date string (YYYY-MM-DD) */
+  /**
+   * The formatted picker date string (YYYY-MM-DD)
+   * The DatePicker returns a Date object with 00:00 UTC. Converting to Pacific
+   * results in an incorrect display date. We need to convert to string without
+   * changing the timezone info in order to get an accurate date.
+  */
   get pickerDate (): string {
     if (!this.date) return ''
 
--- a/src/components/date-picker/DatePicker.vue
+++ b/src/components/date-picker/DatePicker.vue
@@ -17,10 +17,10 @@
       bottom
       min-width="290"
     >
-      <template #activator="{ on }">
+      <template #activator="{ props }">
         <span
           :class="{'date-text-field-pointer': enableSelector}"
-          v-on="enableSelector && on"
+          v-bind="enableSelector && props"
         >
           <v-text-field
             id="date-text-field"
@@ -30,9 +30,9 @@
             :clearable="clearable"
             :error-messages="errorMsg"
             :error="!!errorMsg"
-            :value="displayDate"
+            :model-value="displayDate"
             :label="title"
-            :name="Math.random()"
+            :name="Math.random().toString()"
             :rules="inputRules"
             :disabled="disablePicker"
             :hint="hint"
@@ -41,24 +41,25 @@
             filled
             @click:clear="emitClear()"
             @keydown="$event.preventDefault()"
-            @keyup.enter="emitDate(dateText)"
+            @keyup.enter="emitDate(date)"
           />
         </span>
       </template>
       <v-date-picker
         id="date-picker-calendar"
-        v-model="dateText"
+        v-model="date"
+        color="primary"
         width="490"
         :min="minDate"
         :max="maxDate"
       >
-        <template #default>
+        <template #actions>
           <div>
             <v-btn
               id="btn-done"
               text
               color="primary"
-              @click="emitDate(dateText)"
+              @click="emitDate(date)"
             >
               <strong>OK</strong>
             </v-btn>
@@ -72,20 +73,18 @@
             </v-btn>
           </div>
         </template>
-      </v-date-picker>
+      </v-date-picker> 
     </v-menu>
   </v-form>
 </template>
 
 <script lang="ts">
-import { Component, Emit, mixins, Prop, Watch, Vue } from 'vue-facing-decorator'
+import { Component, Emit, Prop, Watch } from 'vue-facing-decorator'
 import { FormIF } from '@bcrs-shared-components/interfaces'
 import { DateMixin } from '@/mixins' // NB: local mixin (StoryBook can't find it otherwise)
 
-@Component({
-  mixins: [DateMixin]
-})
-export default class DatePicker extends Vue {
+@Component({})
+export default class DatePicker extends DateMixin {
   // Add element types to refs
   $refs!: {
     form: FormIF,
@@ -108,12 +107,12 @@ export default class DatePicker extends Vue {
   @Prop({ default: false }) readonly persistentHint!: boolean
   @Prop({ default: false }) readonly clearable!: boolean
 
-  private dateText = null
+  private date = null
   private displayPicker = false
 
   /** Clear local model after each action. */
   public clearDate (): void {
-    this.dateText = ''
+    this.date = null
     this.displayPicker = false
   }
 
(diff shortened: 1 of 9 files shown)


src/components/date-picker/DatePicker.vue
```diff
@@ -17,10 +17,10 @@
       bottom
       min-width="290"
     >
-      <template #activator="{ on }">
```
leodube-aot, line 20: The `v-menu` would not open with `v-on` and `on`. Using `props` seems to work

src/components/date-picker/DatePicker.vue
```diff
@@ -30,9 +30,9 @@
             :clearable="clearable"
             :error-messages="errorMsg"
             :error="!!errorMsg"
-            :value="displayDate"
+            :model-value="displayDate"
             :label="title"
-            :name="Math.random()"
```
leodube-aot, line 35: `name` expects a string

src/components/date-picker/DatePicker.vue
```diff
@@ -41,24 +41,25 @@
...
             @click:clear="emitClear()"
             @keydown="$event.preventDefault()"
-            @keyup.enter="emitDate(dateText)"
+            @keyup.enter="emitDate(date)"
           />
         </span>
       </template>
       <v-date-picker
         id="date-picker-calendar"
-        v-model="dateText"
```
leodube-aot, line 50: The Vuetify 3 DatePicker does not seem to be working with `string` dates, so I switched the private `dateText` variable to be `data` with type `Date`.

src/mixins/date-mixin.ts
```diff
@@ -65,18 +65,19 @@ export default class DateMixin extends Vue {
   }
 
   /**
-   * Converts a Date object to a date string (YYYY-MM-DD) in Pacific timezone.
+   * Converts a Date object to a date string (YYYY-MM-DD)
+   * @param convertTimezone whether to convert the timzone to Pacific
    * @example "2021-01-01 07:00:00 GMT" -> "2020-12-31"
    * @example "2021-01-01 08:00:00 GMT" -> "2021-01-01"
    */
-  dateToYyyyMmDd (date: Date): IsoDatePacific {
```
leodube-aot, line 72: Since Vuetify 3 Datepicker returns a `Date`, it also includes timezone information and defaults the time to `00:00:00`. Our previous datepicker would ignore this since it would return a string. 

I tested the Vuetify 3 DatePicker for non-pacific timezones and this resulted in bizarre behaviour. Notice below that Dec 4 is selected in the picker, but the text field shows Dec 3. The console.log shows that the time is set to 12AM CST, which means when it converts to a pacific date in this mixin function, it sets the date back a day.

Screenshot 2023-12-06 at 8 51 08 AM

My solution is to update this function to allow the option of ignoring the timezone.

src/components/date-picker/DatePicker.vue
severinbeauvais: Did you see this? Does it help? https://stackoverflow.com/questions/75959657/replacement-for-vue-2-v-on-in-vue-3

src/components/date-picker/DatePicker.vue
```diff
@@ -30,9 +30,9 @@
             :clearable="clearable"
             :error-messages="errorMsg"
             :error="!!errorMsg"
-            :value="displayDate"
+            :model-value="displayDate"
```
severinbeauvais, line 33: This is a two-way binding now, right? So does displayDate get updated when the user types in something?

src/components/date-picker/DatePicker.vue
```diff
@@ -161,11 +163,11 @@ export default class DatePicker extends Vue {
...
   }
 
-  @Watch('dateText')
+  @Watch('date')
   @Emit('emitDateSync')
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  private emitDateSync (date: string): string {
-    return this.dateText
+  private emitDateSync (date: Date): string {
+    return this.date
```
severinbeauvais, line 170: This is a significant change to what was emitted previously, which would require some rework of all UIs that use this shared component.

I recommend emitting the same string value as before.

severinbeauvais: OK, I'm a bit confused -- what's the difference between emitDate() above and emitDateSync() here -- why do we need both?

src/components/date-picker/package.json
```diff
@@ -6,10 +6,9 @@
   },
   "dependencies": {
     "@bcrs-shared-components/interfaces": "^1.0.76",
-    "@bcrs-shared-components/mixins": "^1.1.27",
-    "vue": "^2.7.14"
+    "@bcrs-shared-components/mixins": "^1.1.27"
```
severinbeauvais, line 9: I think you don't actually import the shared mixins. Line 84 in the component imports the "local mixin" -- a relative path from this component. (`@/mixins` always resolves to the local mixins folder.)

Basically, when this component is used in Storybook, it uses DateMixin in this repo's mixins folder, but when this component is used in an app, it uses DateMixin in the app's mixins folder.

See if you can get this working while importing the shared mixin instead. 🤞

src/mixins/date-mixin.ts
severinbeauvais: Two things:
1. I'd rather not change the date mixin.
2. When this component is used in an app, it's using the app's date mixin (see comment above).

If this method isn't working for you then it means the `date` parameter may be invalid. Normally, a date object is UTC/GMT internally and has a tz for display purposes. Try some of these on `date` and see how it's set internally.

image

As for this method, it should always convert a date into America/Vancouver time zone.

src/components/date-picker/DatePicker.vue
leodube-aot: Good point, I'll emit the string instead of the Date. I'm not too sure what the point of emitDateSync() is. I'll dig through some of the UIs a bit to see if it's being used.

src/components/date-picker/DatePicker.vue
leodube-aot: It doesn't update since the field is readonly. The picker wasn't updating when it was `value`, I think because in Vue 3 they changed that prop to `model-value`: https://v3-migration.vuejs.org/breaking-changes/v-model.html

src/components/date-picker/DatePicker.vue
leodube-aot: It seems it's used in [Edit UI](https://github.com/bcgov/business-edit-ui/blob/main/src/dialogs/ResolutionDateDialog.vue#L22) and a [few places in Create UI](https://github.com/search?q=repo%3Abcgov%2Fbusiness-create-ui+emitDateSync&type=code)

src/components/date-picker/package.json
leodube-aot: Tried importing the shared mixin and the component seems to work fine. I'll include in my next commit.

src/components/date-picker/DatePicker.vue
leodube-aot: The way I've seen it done in a couple of Ketaki's tickets is to use kebab-case, I think because we generally use it for our props already

src/mixins/date-mixin.ts
```diff
@@ -65,7 +65,7 @@ export default class DateMixin extends Vue {
   }
 
   /**
-   * Converts a Date object to a date string (YYYY-MM-DD) in Pacific timezone.
```
severinbeauvais, line 68: This is key information. Please restore it.

(The resulting yyyy-mm-dd may be different in a different timezone.)

src/components/date-picker/DatePicker.vue
```diff
@@ -144,9 +144,10 @@ export default class DatePicker extends Vue {
 
   /** Emit date to add or remove. */
   @Emit('emitDate')
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  protected emitDate (date: string): void {
+  protected emitDate (): string {
     this.displayPicker = false
+    // Convert to ISO string and truncate to just date portion
+    return this.date?.toISOString()?.slice(0, 10)
```
severinbeauvais, line 150: I think you should use the DateMixin (or DateUtilities) function to convert this.

The problem is that a UTC/GMT/ISO string is 7 or 8 hours ahead of Pacific time, so you will get the wrong value here for the first 7 or 8 hours of a day (it will be ahead by 1 day).

src/components/date-picker/DatePicker.vue
severinbeauvais: So both are needed? Please resolve if no change is needed.

severinbeauvais:
> > Does it work if you import the "local" mixin instead of the shared one?
> 
> No unfortunately it doesn't

OK, that's partly good news. (It means the previous issue was resolved.)

What if you add the shared mixin to Storybook's (ie, this repo's) package file?

severinbeauvais:
> > What if you add the shared mixin to Storybook's (ie, this repo's) package file?
> 
> It looks like the shared mixin is already a dependency of the base package.json

Oh yeah. So maybe it's a Storybook build/config issue? How does Storybook build this so that DatePicker can see the mixins?

@ketaki-deodhar , have you run across this yet?

ketaki-deodhar:
> > > What if you add the shared mixin to Storybook's (ie, this repo's) package file?
> > 
> > 
> > It looks like the shared mixin is already a dependency of the base package.json
> 
> Oh yeah. So maybe it's a Storybook build/config issue? How does Storybook build this so that DatePicker can see the mixins?
> 
> @ketaki-deodhar , have you run across this yet?

@severinbeauvais not really. Let me take a look

src/components/date-picker/DatePicker.vue
leodube-aot: You're right, it would've been the wrong value early in the day. I've replaced the `toISOString` calls I used with functions from DateMixin. However, using the mixin still results in the previous problem of the displayed date being wrong since the picker defaults to 00:00:00 GMT time which, when converted to PST by the mixin, puts the displayed value back a day.

I think the previous solution of using `toLocaleDateString` without converting the timezone would work, but I'm not sure if it should go in the util or in the component itself.

leodube-aot:
> can you try passing the mixin in the mount? Not sure if it will work or if right way to do it but something like below..... !

No luck unfortunately. I passed the mixin to the mount like so:

Screenshot 2023-12-07 at 2 55 02 PM

I tried it with the previous method of adding the mixin to the component, which doesn't seem to register the mixin based on the error message. In this case, both storybook and the unit tests fail.

Screenshot 2023-12-07 at 2 52 44 PM

I also tried it with the way I have it in this PR. In this case storybook works but the unit tests fail.

Screenshot 2023-12-07 at 2 53 41 PM

src/components/date-picker/DatePicker.vue
severinbeauvais: Oh, I didn't realize the picker sets it to 0Z :(

As this is specific to this component, I think you should work around it in here. My suggestion is to create a computed (local getter) in here to provide a yyyy-mm-dd that you expect (ie, yes, truncate the ISO date if that's easiest). Then you can use this computed to emit the events.

leodube-aot: I've fixed it following this approach of using a local getter and I think it it works pretty well, even if it's partially duplicated from the mixin.

leodube-aot:
> OK, you got it down to a unit test error. That's progress. And if we decide to give up, we can forgo the unit tests for now (but we'd create a Tech Debt ticket to fix them later).

It also fails before any of the tests are run (see the test summary copied below, I'm running the tests only for DatePicker.spec.ts so that's why there's only one test file). I imagine it might be some issue with the test setup that's thrown from the component. One thing I'm confused by is the `mount`ing of the component only occurs within the tests themselves, but based on the output below none of them get a chance to run. 

```
Test Files  1 failed (1)
Tests  no tests
Start at  15:14:26
Duration  2.06s
```

src/components/date-picker/DatePicker.vue
```diff
@@ -127,9 +127,25 @@ export default class DatePicker extends DateMixin {
     this.date = this.yyyyMmDdToDate(this.initialValue)
   }
 
+  /** The formatted picker date string (YYYY-MM-DD) */
+  get pickerDate (): string {
```
severinbeauvais, line 131: Please add a comment indicating you need to do this because the picker returns a Date object with 00:00 UTC.

src/components/date-picker/DatePicker.vue
severinbeauvais: Looks good to me. Pretty clean!

src/components/date-picker/DatePicker.vue
severinbeauvais: Perfect 👍

severinbeauvais (approved):

leodube-aot:
Created new techdebt ticket for fixing up the unit tests. I'll get one more review before merging this PR.

https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/gh/bcgov/entity/18943

ketaki-deodhar (approved):
Looks good!

ketaki-deodhar:
> > can you try passing the mixin in the mount? Not sure if it will work or if right way to do it but something like below..... !
> 
> No luck unfortunately. I passed the mixin to the mount in the unit tests like so:
> 
> Screenshot 2023-12-07 at 2 55 02 PM
> I ran it with the previous method of adding the mixin to the component, which doesn't seem to register the mixin based on the error message. In this case, both storybook and the unit tests fail.
> 
> Screenshot 2023-12-07 at 2 52 44 PM
> I also tried running it with the way I extend the mixin in this PR. In this case storybook works but the unit tests fail.
> 
> Screenshot 2023-12-07 at 2 53 41 PM

Unit tests are a big pain point :(

leodube-aot:
> Unit tests are a big pain point :(

I agree haha I felt like I was going in circles yesterday

PR #216, opened 2023-12-07
17086 update shareStructure.vue

tshyun24:
*Issue #:* /bcgov/entity#17086

*Description of changes:* update v-data-table to have the shareStructure.vue have the same data table as before, but "correct" and "change" button in the table does not work, because it calls another editShareStructure.vue file.


src/components/share-structure/ShareStructure.vue
```diff
@@ -38,7 +38,7 @@
       <div class="btn-container py-6 px-4">
         <v-btn
           id="btn-add-person"
-          outlined
+          variant="outlined"
```
tshyun24, line 41: since `flat / outlined / text / plain props have been combined into a single variant prop.`
https://vuetifyjs.com/en/getting-started/upgrade-guide/#v-btn-v-btn-toggle

src/components/share-structure/ShareStructure.vue
```diff
@@ -73,7 +73,7 @@
     </div>
 
     <v-data-table
-      class="share-structure-table"
+      headerProps="share-structure-table"
```
tshyun24, line 76: since `class has been replaced with headerProps.`
https://vuetifyjs.com/en/getting-started/upgrade-guide/#v-data-table

src/components/share-structure/ShareStructure.vue
```diff
@@ -179,13 +179,13 @@
                   offset-y
                   left
                 >
-                  <template #activator="{ on }">
+                  <template #activator="{ props }">
```
tshyun24, line 182: since `Activator slots work slightly different. Replace #activator={ attrs, on } with #activator={ props }, then remove v-on="on" and replace v-bind="attrs" with v-bind="props"`
https://vuetifyjs.com/en/getting-started/upgrade-guide/#general-changes

src/plugins/vuetify.ts
```diff
@@ -17,6 +18,9 @@ export default createVuetify({
       mdi,
     },
   },
-  components,
+  components: {
+    ...components,
+    VDataTable
+  },
```
tshyun24, line 24: add this, otherwise `v-data-table` component cannot be founded
https://stackoverflow.com/questions/65669486/v-data-table-component-not-found

tshyun24:
image
looks like that right now

src/plugins/vuetify.ts
tshyun24: remove that one and with the updated Vuetify version, it works

severinbeauvais: Great news! I think you can resolve this conversation then :)

src/components/share-structure/ShareStructure.vue
```diff
@@ -1187,24 +1187,30 @@ export default class ShareStructure extends Vue {
   opacity: .4;
 }
 
+.v-btn {
+  text-transform: none !important;
+}
```
severinbeauvais, line 1192: Why do you need this? Please add a code comment.

src/components/share-structure/package.json
```diff
@@ -10,10 +10,9 @@
     "@bcrs-shared-components/enums": "^1.0.51",
     "@bcrs-shared-components/interfaces": "^1.0.76",
     "lodash": "4.17.21",
-    "uuid": "^9.0.0",
-    "vue": "^2.7.14"
```
severinbeauvais, line 14: I think EditShareStructure, which is a part of this component, still imports Vue.

severinbeauvais (approved):
A couple of small comments but otherwise LGTM. Please get another review approval or 2.

I know yout ticket was to handle the v-data-table only, but while you're here, does this Share Structure component work properly under Vue3? If not, what else needs to be updated?

leodube-aot (approved):
LGTM!

src/components/share-structure/ShareStructure.vue
severinbeauvais: Are you saying all VButtons now display their label in uppercase? 🤦 

This will affect ALL our buttons. Instead of overriding all of them individually, I think we should add the override in the global assets folder (in override file). Thoughts?

tshyun24:
> I know ~yout~ **your** ticket was to handle the v-data-table only, but while you're here, does this Share Structure component work properly under Vue3? If not, what else needs to be updated?

Yeah, so for now, the dropdown bar works for `move up` and `move down`,  `add series` will request the update at `EditShareStructure.vue`, and 'remove' need some modifications at https://github.com/bcgov/bcrs-shared-components/blob/6ac54d98102ec0a8d91beabccf33716d724d2998/src/components/share-structure/ShareStructure.vue#L1007
image.

And also `Add Share Class` button does not work, also request the update at `EditShareStructure.vue`.

severinbeauvais:
Shaoyun, please create a new ticket (in the same epic) to do the remaining work you identified above to get this component completely functional in Vue 3. Please post the new ticket number here. Thanks!

src/components/share-structure/package.json
```diff
@@ -10,7 +10,8 @@
     "@bcrs-shared-components/enums": "^1.0.51",
     "@bcrs-shared-components/interfaces": "^1.0.76",
     "lodash": "4.17.21",
-    "uuid": "^9.0.0"
+    "uuid": "^9.0.0",
+    "vue": "^2.7.14"
```
severinbeauvais, line 14: This should be Vue3 now, right?

tshyun24: yesyes!!

tshyun24:
> Shaoyun, please create a new ticket (in the same epic) to do the remaining work you identified above to get this component completely functional in Vue 3. Please post the new ticket number here. Thanks!

Hiii Sev, should it be on the new Zenhub board or the one that we are using now?

src/components/share-structure/ShareStructure.vue
severinbeauvais: Shaoyun, what are you thoughts on this?

severinbeauvais:
> Hiii Sev, should it be on the new Zenhub board or the one that we are using now?

We're still using the old board.

src/components/share-structure/ShareStructure.vue
tshyun24: > Are you saying all VButtons now display their label in uppercase? 🤦
> 
> This will affect ALL our buttons. Instead of overriding all of them individually, I think we should add the override in the global assets folder (in override file). Thoughts?

sorry for missing this messagekkk. looking it now!

tshyun24:
New ticket for the remaining stuff:
https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/gh/bcgov/entity/18939

src/assets/styles/overrides.scss
```diff
@@ -7,7 +7,8 @@
   text-transform: none;
   font-weight: 500;
   letter-spacing: normal;
-
+  text-transform: none !important; //otherwise the label will be uppercase
```
severinbeauvais, line 10: Oh, so this replaces line 7?

tshyun24: yeah... miss line 7, updated with adding "important"

PR #218, opened 2023-12-11
18974+18950 Minor InputFieldDatePicker Updates

cameron-eyds:
*Issue #:* /bcgov/entity#18950

*Description of changes:*
* Updates min/max date prop values to child component (baseDatePick) to only set when passed. 
* Minor clean up of functions
* Update date format emit. eg from `Dec 11, 2023` to `2023-12-11` to fit current implementations. 
- Note: The above was adhere to our current handling, we format all display dates from the latter and it also is an acceptable Date format for our api schemas. 

*ROUTER UPDATE NOTE:*
- I've upgraded the shared components vue-router version to vue-router4. This is the version you will want for vue3. 
- It has the following functions available `useRoute`, `useRouter`.
- I didn't see this used in any of the other components but worth looking into if this affects the storybook build?

Screenshot 2023-12-11 at 2 42 04 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "sbc-common-components": "4.0.1",
     "uuid": "^9.0.0",
     "vue": "^3.3.4",
-    "vue-router": "^3.6.5",
+    "vue-router": "^4.2.5",
     "vue-the-mask": "^0.11.1",
     "@vuelidate/core": "^2.0.3",
     "@vuelidate/validators": "^2.0.4",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bcrs-shared-components",
-  "version": "3.0.4",
+  "version": "3.0.5",
   "private": true,
   "appName": "BCRS Shared Components",
   "scripts": {
--- a/src/utils/date-helper.ts
+++ b/src/utils/date-helper.ts
@@ -1,6 +1,8 @@
 import { isDate, isEmpty } from 'lodash'
 // Converts date to string and pacific date string
 // Example Output: August 11, 2023
+import { isDate } from 'lodash'
+
 export function shortPacificDate (date: Date | string): string {
   // safety check
   if (!isDate(date) || isNaN(date.getTime())) return null
@@ -9,4 +11,21 @@ export function shortPacificDate (date: Date | string): string {
     dateStyle: 'long',
     timeZone: 'America/Vancouver'
   }).format(new Date(date))
+}
+
+/**
+ * Converts a Date object to a date string (YYYY-MM-DD) in Pacific timezone.
+ * @example "2021-01-01 07:00:00 GMT" -> "2020-12-31"
+ * @example "2021-01-01 08:00:00 GMT" -> "2021-01-01"
+ */
+export function dateToYyyyMmDd (date: Date): string {
+  // safety check
+  if (!isDate(date) || isNaN(date.getTime())) return null
+
+  return date.toLocaleDateString('en-CA', {
+    timeZone: 'America/Vancouver',
+    month: 'numeric', // 12
+    day: 'numeric', // 31
+    year: 'numeric' // 2020
+  })
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "bcrs-shared-components",
-  "version": "3.0.4",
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "sbc-common-components": "4.0.1",
     "uuid": "^9.0.0",
     "vue": "^3.3.4",
-    "vue-router": "^3.6.5",
+    "vue-router": "^4.2.5",
```
severinbeauvais, line 26: Ketaki, Leo - Cameron was kind enough to test this component in his (ppr/mhr) Vue UI, but needed these changes. He's swamped so I said we'd take over these changes. Please review. Who has time to continue this PR?

Since this file has been updated, its app version should be updated, too.

src/utils/date-helper.ts
```diff
@@ -1,8 +1,27 @@
...
+  // safety check
+  if (!isDate(date) || isNaN(date.getTime())) return null
+
+  return date.toLocaleDateString('en-CA', {
+    timeZone: 'America/Vancouver',
+    month: 'numeric', // 12
+    day: 'numeric', // 31
+    year: 'numeric' // 2020
+  })
+}
```
severinbeauvais, line 27: Should use this method from date mixin instead.

severinbeauvais: And actually... instead of using a mixin for date methods, we should have a basic "date utilities" class like here: https://github.com/bcgov/business-filings-ui/blob/main/src/services/date-utilities.ts

It could be under utils/ but I think utils isn't exported, so maybe having it in services/ is a better idea.

Thoughts?

package.json
cameron-eyds: Will do!

src/utils/date-helper.ts
cameron-eyds: I don't believe it needs to be a class, utility functions could be stand-alone. 
I feel it's a bit more portable than a class or mixin but if vue3 converted components are going to remain class-based, than perhaps it would be easier to extend as a class or mixin?

Suppose it boils down to which vue api the project leverages :)

package.json
severinbeauvais: Cam, we can do that.

cameron-eyds: Ok great, sounds good!

src/utils/date-helper.ts
severinbeauvais: I'd like to keep all date methods in some component (class or mixin) that all UIs could use, ie, make it shared.

There's already a shared date mixin, but I'd like to make THIS component not depend on it, so let's just leave alone for now. It's a bit of tech debt unfortunately but it's a lot of churn the other way.

cameron-eyds: Definitely a good point!
Another option we could explore is making a plugin, if the goal was making portable functionality. 
If a plugin was defined and exported with the desired mixins/functionality, it might be barely more than a wrapper per se but there would be benefits on the project side. 

We could include said plugin at the root level of a project, making the functionality available globally without the need to extending or instantiating at the component level.

Anywho, i'm just trailing off down a rabbit hole, this is just theoretical haha.

package.json
severinbeauvais: Thank you so much! Really appreciate this.

Can you reuse your prev ticket number or do you want a new 2-pointer for this?

leodube-aot: No problem! I think we should create a new 2-pointer for this.

severinbeauvais: Make it so :)

leodube-aot: Ticket for completing this PR: https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/gh/bcgov/entity/18974

leodube-aot:
- Updated the package version. 
- I checked out the components that we've currently updated to Vue 3 and the vue-router upgrade does not affect them. 
- I had a look into creating unit tests for this component, but until [the issue with v-menu in unit tests](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/gh/bcgov/entity/18943) is resolved they cannot be created for any of the date picker components.

severinbeauvais (approved):
This is fine by me!

cameron-eyds:
I put a small comment in ticket 18943 which hopefully can help with the unit testing aspect of the overlays.

PR #220, opened 2023-12-13
18777 - Vue3 upgrade: enums, interfaces, mixins PART 2

ketaki-deodhar:
*Issue #:* /bcgov/entity#18777

*Description of changes:*

This is continuation of previous [PR](https://github.com/bcgov/bcrs-shared-components/pull/217) for unit tests

- services: unit tests for `fee-services`
- unit tests for `date-helper`

### DIFF
--- a/src/utils/date-helper.ts
+++ b/src/utils/date-helper.ts
@@ -1,6 +1,10 @@
+import { isDate, isEmpty } from 'lodash'
 // Converts date to string and pacific date string
 // Example Output: August 11, 2023
 export function shortPacificDate (date: Date | string): string {
+  // safety check
+  if (!isDate(date) || isNaN(date.getTime())) return null
+
   return new Intl.DateTimeFormat('en-US', {
     dateStyle: 'long',
     timeZone: 'America/Vancouver'
--- a/src/utils/date-helper.ts
+++ b/src/utils/date-helper.ts
@@ -1,0 +1,8 @@
+// Converts date to string and pacific date string
+// Example Output: August 11, 2023
+export function shortPacificDate (date: Date | string): string {
+  return new Intl.DateTimeFormat('en-US', {
+    dateStyle: 'long',
+    timeZone: 'America/Vancouver'
+  }).format(new Date(date))
+}
--- a/tests/unit/date-helper.spec.ts
+++ b/tests/unit/date-helper.spec.ts
@@ -2,10 +2,10 @@ import { shortPacificDate } from '../../src/utils/date-helper'
 
 describe('Date Helper', () => {
   it('returns short pacific date shortPacificDate()', () => {
-    expect(shortPacificDate('2021-07-01T00:00:00+00:00')).toBe('June 30, 2021')
+    expect(shortPacificDate(new Date('2021-06-30 07:00:00 GMT'))).toBe('June 30, 2021')
   })
 
   it('returns RangeError when empty string is passed shortPacificDate()', () => {
-    expect(() => shortPacificDate('')).toThrowError('Invalid time value')
+    expect(shortPacificDate('')).toBe(null)
   })
 })
--- a/tests/unit/date-helper.spec.ts
+++ b/tests/unit/date-helper.spec.ts
@@ -5,7 +5,7 @@ describe('Date Helper', () => {
     expect(shortPacificDate(new Date('2021-06-30 07:00:00 GMT'))).toBe('June 30, 2021')
   })
 
-  it('returns RangeError when empty string is passed shortPacificDate()', () => {
+  it('returns null when empty string is passed shortPacificDate()', () => {
     expect(shortPacificDate('')).toBe(null)
   })
 })
--- a/tests/unit/date-helper.spec.ts
+++ b/tests/unit/date-helper.spec.ts
@@ -1,0 +1,11 @@
+import { shortPacificDate } from '../../src/utils/date-helper'
+
+describe('Date Helper', () => {
+  it('returns short pacific date shortPacificDate()', () => {
+    expect(shortPacificDate('2021-07-01T00:00:00+00:00')).toBe('June 30, 2021')
+  })
+
+  it('returns RangeError when empty string is passed shortPacificDate()', () => {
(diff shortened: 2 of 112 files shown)


tests/unit/fee-services.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+  const filingCodes = [{
+    filingDescription: 'Annual Filing',
+    filingTypeCode: 'OTANN',
+    waiveFees: false,
+    entityType: 'CP',
+    priority: false,
+    futureEffective: false }]
+
+  beforeAll(() => {
+    vi.mocked(axios.get).mockClear()
```
ketaki-deodhar, line 27: Ref: https://vitest.dev/api/vi.html#vi-mock

tests/unit/date-helper.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { shortPacificDate } from '../../src/utils/date-helper'
+
+describe('Date Helper', () => {
+  it('returns short pacific date shortPacificDate()', () => {
+    expect(shortPacificDate('2021-07-01T00:00:00+00:00')).toBe('June 30, 2021')
+  })
+
+  it('returns RangeError when empty string is passed shortPacificDate()', () => {
+    expect(() => shortPacificDate('')).toThrowError('Invalid time value')
```
severinbeauvais, line 9: Do you think it would be better if the method returned null instead? (Do its user expect it to throw on error?)

tests/unit/fee-services.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+import axios from 'axios'
+import FeeServices from '../../src/services/fee-services'
+
+vi.mock('axios', () => ({
+  get: vi.fn(),
+  all: vi.fn(),
+  spread: vi.fn()
+}))
+
+vi.mock('axios')
```
severinbeauvais, line 11: What does this do (compared with line 5)?

tests/unit/fee-services.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    vi.mocked(axios.spread).mockReturnValue(mockAxiosSpreadResult)
+    FeeServices.getFee(filingCodes, API_URL)
+  })
+
+  it('calls axios.get once with extra fee parameters ', () => {
+    expect(axios.get)
+      .toHaveBeenCalledWith(`${API_URL}fees/BC/BCRSF?priority=true&futureEffective=true`,
+        { 'headers': { 'Account-Id': 0, 'Authorization': 'Bearer null' } })
+  })
+})
```
severinbeauvais, line 93: Please add tests where you mock the response and verify that `getFee()` returns the expected fee array.

Maybe also add some tests for `prepareUrl()`.

These don't need to be elaborate -- just enough to verify that the service methods are doing what we expect of them. (I'm OK with testing the go path and maybe a single error path.)

tests/unit/date-helper.spec.ts
ketaki-deodhar: I did try that but get this. 

image

I am now thinking it might be the way function is written is the casue

severinbeauvais: Yes. I was thinking that the function should be updated to handle this gracefully instead of throwing an exception. Thoughts?

ketaki-deodhar: Yes function needs to be updated.

tests/unit/fee-services.spec.ts
ketaki-deodhar: Thanks for spotting that. With this default mock (`vi.mock'), code on line 5 is not needed. I will remove it.

JazzarKarim (approved):
No comments from me Ketaki. LGTM

src/utils/date-helper.ts
```diff
@@ -1,6 +1,10 @@
+import { isDate, isEmpty } from 'lodash'
 // Converts date to string and pacific date string
 // Example Output: August 11, 2023
 export function shortPacificDate (date: Date | string): string {
+  // safety check
+  if (!isDate(date) || isNaN(date.getTime())) return null
```
ketaki-deodhar, line 6: added a safety check

tests/unit/fee-services.spec.ts
ketaki-deodhar: Removed

tests/unit/fee-services.spec.ts
ketaki-deodhar: After 'struggling in vain',  I have decided to create a `techdebt` ticket to add more unit tests for `axios` call. I will add the details in it.

src/utils/date-helper.ts
severinbeauvais: Will this fail if date is passed in as an empty string?

severinbeauvais: I see you just verified that in the unit test below 😛

tests/unit/date-helper.spec.ts
```diff
@@ -2,10 +2,10 @@ import { shortPacificDate } from '../../src/utils/date-helper'
...
 describe('Date Helper', () => {
   it('returns short pacific date shortPacificDate()', () => {
-    expect(shortPacificDate('2021-07-01T00:00:00+00:00')).toBe('June 30, 2021')
+    expect(shortPacificDate(new Date('2021-06-30 07:00:00 GMT'))).toBe('June 30, 2021')
   })
 
   it('returns RangeError when empty string is passed shortPacificDate()', () => {
-    expect(() => shortPacificDate('')).toThrowError('Invalid time value')
+    expect(shortPacificDate('')).toBe(null)
   })
```
severinbeauvais, line 10: Please fix the unit test label above.

Also, please add a unit test for an invalid-format date (like a string that isn't a date).

severinbeauvais (approved):
GTG after you handle my last comment. (I don't need to re-review this.)

tests/unit/date-helper.spec.ts
ketaki-deodhar: Done

PR #225, opened 2023-12-22
18788 Added correct-name sub-components for AML

severinbeauvais:
*Issue #:* bcgov/entity#18788

*Description of changes:*
- added CorrectAmlAdopt sub-component
- added CorrectAmlNumbered sub-component
- updated whitespace in CorrectName
- added new sub-components, prop and options  to CorrectName
- removed unneeded layout in CorrectNameToNumber
- added optional legal type check to validateNameRequest()

### DIFF
--- a/src/components/correct-name/CorrectName.vue
+++ b/src/components/correct-name/CorrectName.vue
@@ -5,7 +5,7 @@
   >
     <p
       v-if="!isOneOption"
-      class="info-text mb-5 pb-5 bottom-border"
+      class="info-text mb-0 pb-5 bottom-border"
     >
       You can {{ actionTxt || 'correct the company name' }} in one of the following ways:
     </p>
@@ -19,7 +19,7 @@
         v-for="(item,i) in displayedOptions"
         :id="`x-panel-${item.id}`"
         :key="i"
-        class="mb-4"
+        class="py-4"
         :disabled="isOneOption"
         @click="identifyForm(item.id)"
       >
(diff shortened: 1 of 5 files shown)


src/components/correct-name/CorrectName.vue
severinbeauvais, line 1: Sample new layout (whitespace):

image

src/components/correct-name/CorrectNameRequest.vue
```diff
@@ -26,6 +26,7 @@
         <v-text-field
           id="nr-number"
           v-model="nrNumber"
+          autofocus
```
severinbeauvais, line 29: We need to be careful when using this (because too many doesn't work well), but in specific cases this brings focus to the text field, which is very nice.

src/components/correct-name/CorrectNameToNumber.vue
```diff
@@ -4,17 +4,13 @@
     ref="form"
     v-model="formValid"
   >
-    <v-row no-gutters>
-      <v-col>
```
severinbeauvais, line 8: A single row with a single column is a lazy way to lay out a flex panel, and in fact it's not needed here.

image

src/components/correct-name/CorrectNameRequest.vue
ketaki-deodhar: Good to know!

ketaki-deodhar (approved):
Looks good

PR #230, opened 2024-01-03
18992 - Limited Restoration Panel Vue 3 upgrade

ketaki-deodhar:
*Issue #:* /bcgov/entity#18992

*Description of changes:*

- Limited Restoration Panel Vue 3 upgrade
- update unit tests

### DIFF
--- a/src/components/limited-restoration-panel/LimitedRestorationPanel.vue
+++ b/src/components/limited-restoration-panel/LimitedRestorationPanel.vue
@@ -6,33 +6,36 @@
   >
     <v-radio
       id="radio-24"
-      class="radio-button pt-2"
+      class="pt-2"
+      color="#1976d2"
       label="2 years"
       value="24"
     />
     <v-radio
       id="radio-18"
-      class="radio-button pt-2"
+      color="#1976d2"
       label="18 months"
       value="18"
     />
     <v-radio
       id="radio-12"
-      class="radio-button pt-2"
+      color="#1976d2"
       label="12 months"
       value="12"
     />
     <v-radio
       id="radio-6"
-      class="radio-button pt-2"
+      color="#1976d2"
       label="6 months"
       value="6"
     />
-    <v-row class="ml-0 mt-0 radio-button pt-2">
+    <v-row class="ml-0 mt-0 pt-2">
       <v-radio
         id="radio-custom"
         class="mt-n4"
+        color="#1976d2"
         value="customMonths"
+        inline
       />
       <v-form ref="monthsRef">
         <v-text-field
@@ -40,12 +43,12 @@
           v-model="inputValue"
           class="text-field-months"
           type="number"
-          dense
-          filled
+          density="compact"
+          variant="filled"
           hide-spin-buttons
           :rules="monthsRules"
           :disabled="(radioValue !== 'customMonths')"
-          @input="onMonthsInput($event)"
+          @update:model-value="onMonthsInput($event)"
         />
       </v-form>
       <span class="ml-2 mt-2 month-text">month(s)</span>
@@ -160,10 +163,9 @@ export default class LimitedRestorationPanel extends Vue {
   // decrease the month text field's width
   .text-field-months {
     width: 3.5rem;
-    .error--text {
-      margin-left: -0.5rem;
-      position: absolute;
-      width: 10rem;
+    .v-input__details{
+      width: 200px;
+      padding-left: 2px;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -6,46 +6,49 @@
   >
     <v-radio
       id="radio-24"
-      class="radio-button pt-2"
+      class="pt-2"
+      color="#1976d2"
```
severinbeauvais, line 10: We should not have hex colours in code.

Should this be "primary" instead?

If this is a very special case (but I didn't see it in Leo's PR with radio buttons), since it isn't easy to add a new colour to both this repo and the apps that use this component, I'd be OK with using `:color="myNamedColor"` here and declaring myNamedColor in the code below.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
```diff
@@ -6,46 +6,49 @@
...
           type="number"
-          dense
-          filled
+          density="compact"
+          variant="filled"
           hide-spin-buttons
           :rules="monthsRules"
           :disabled="(radioValue !== 'customMonths')"
-          @input="onMonthsInput($event)"
+          @update:model-value="onMonthsInput($event)"
```
severinbeauvais, line 51: Interesting. So this works? I wonder why it's different in Leo's component. Can you have a look there please?

tests/unit/LimitedRestorationPanel.spec.ts
```diff
@@ -1,102 +1,133 @@
...
+
+    await wrapper.find('#radio-custom').setChecked()
+    await wrapper.find('#text-field-months').setValue('25')
+
+    expect(wrapper.vm.valid).toBeFalsy()
+
+    wrapper.unmount()
+  })
+
+  it.skip('emits events when we select a preset expiry (24 months)', async () => {
```
severinbeauvais, line 86: What's the plan to fix this unit test (and the others below)?

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
ketaki-deodhar: Yes I had to add this specifically as the buttons were grey colored. I will change as you suggested

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
ketaki-deodhar: Sure.  I just left a comment on that PR

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
severinbeauvais: Will the radio buttons in other components also need fixing?

Does this work correctly if this component is imported into one of our UIs (in which we declare a theme color "primary")?

I am not yet convinced that this is a special enough case to locally declare the color value (instead of using the base CSS variables).

tests/unit/LimitedRestorationPanel.spec.ts
ketaki-deodhar: Frankly no plan as of now. I will create a tech debt ticket for this. May be sometime in near future will have to have heads down time to figure out all unit test fixes.

severinbeauvais: Ok by me.

src/components/limited-restoration-panel/LimitedRestorationPanel.vue
ketaki-deodhar: Looks like this will be an issue for other components as well. I have Leo's PR running locally and it has the same issue.
DEV:
image

LOCAL:
image

severinbeauvais: What do you get if you use a color variable name here, like "primary"?

ketaki-deodhar: I tried with `color="primary"` and it works

tests/unit/LimitedRestorationPanel.spec.ts
severinbeauvais: Ketaki, please resolve this conversation after you've created the new ticket. Thx!

severinbeauvais (approved):

JazzarKarim (approved):

tests/unit/LimitedRestorationPanel.spec.ts
ketaki-deodhar: Unit test fix ticket [19149](https://app.zenhub.com/workspaces/entities-team-space---new-65550db1fcfd2f002838a52a/issues/gh/bcgov/entity/19149)

severinbeauvais:
```
lerna notice
Successfully published:
 - @bcrs-shared-components/limited-restoration-panel@3.0.2
lerna success published 1 package
```

PR #231, opened 2024-01-04
18999 - Vue 3 upgrade: StaffPayment, FolioNumberInput, SubMenu

leodube-aot:
*Issue #:* /bcgov/entity#18999

*Description of changes:*
- Updated StaffPayment component to Vue 3
- Updated FolioNumberInput since it is used in the StaffPayment component
- Updated SubMenu since storybook (for some reason) kept showing errors from that component in the StaffPayment component storybook.

**Note:** I am off until Jan 15

### DIFF
--- a/src/components/folio-number-input/FolioNumberInput.stories.ts
+++ b/src/components/folio-number-input/FolioNumberInput.stories.ts
@@ -1,4 +1,4 @@
-import type { Meta } from '@storybook/vue'
+import type { Meta } from '@storybook/vue3'
 import { FolioNumberInput } from './index'
 
 const meta: Meta<typeof FolioNumberInput> = {
@@ -6,10 +6,12 @@ const meta: Meta<typeof FolioNumberInput> = {
 }
 export default meta
 
-const Template = (args, { argTypes }) => ({
-  props: Object.keys(argTypes),
+const Template = (args) => ({
   components: { FolioNumberInput },
-  template: '<folio-number-input v-bind="$props" />' // $props comes from args below
+  setup () {
+    return { args }
+  },
+  template: '<folio-number-input v-bind="args" />'
 })
 
 export const Default = Template.bind({})
--- a/src/components/folio-number-input/FolioNumberInput.stories.ts
+++ b/src/components/folio-number-input/FolioNumberInput.stories.ts
@@ -16,5 +16,6 @@ const Template = (args) => ({
 
 export const Default = Template.bind({})
 Default['args'] = {
-  folioNumber: '01234567890123456789012345678901234567890123456789ABC@123/*!$ ...'
+  folioNumber: '01234567890123456789012345678901234567890123456789ABC@123/*!$ ...',
+  validate: false
 }
--- a/src/components/folio-number-input/FolioNumberInput.vue
+++ b/src/components/folio-number-input/FolioNumberInput.vue
@@ -6,13 +6,14 @@
   >
     <v-text-field
       id="folio-number-textfield"
-      v-model="folioNumberString"
+      :modelValue="folioNumberString"
       variant="filled"
       label="Folio Number (Optional)"
       :rules="folioNumberRules"
       :disabled="disabled"
       autocomplete="chrome-off"
       :name="Math.random().toString()"
+      @update:folioNumberString="emitFolioNumber($event)"
       @focus="emitFocus($event)"
     />
   </v-form>
@@ -60,12 +61,6 @@ export default class FolioNumberInput extends Vue {
   @Watch('folioNumber')
   private onFolioNumberChange (): void {
     this.folioNumberString = this.folioNumber
-  }
-
-  /** Emit folio number on change. */
-  @Watch('folioNumberString')
-  private onFolioNumberStringChange (): void {
-    this.emitFolioNumber(this.folioNumberString)
   }
 
   /** Prompt the field validations. */
--- a/src/components/folio-number-input/FolioNumberInput.vue
+++ b/src/components/folio-number-input/FolioNumberInput.vue
@@ -6,14 +6,14 @@
   >
     <v-text-field
       id="folio-number-textfield"
-      filled
+      variant="filled"
       label="Folio Number (Optional)"
-      :value="folioNumber"
+      :model-value="folioNumber"
       :rules="folioNumberRules"
       :disabled="disabled"
       autocomplete="chrome-off"
-      :name="Math.random()"
-      @input="emitFolioNumber($event)"
+      :name="Math.random().toString()"
+      @update:model-value="emitFolioNumber($event)"
       @focus="emitFocus($event)"
     />
   </v-form>
@@ -97,7 +97,7 @@ export default class FolioNumberInput extends Vue {
    * @returns True if form is valid and False if not
    */
   public validateFolioNumber (): boolean {
-    return this.$refs.folioForm.validate()
+    return this.$refs.folioForm?.validate()
   }
 }
 </script>
--- a/src/components/folio-number-input/FolioNumberInput.vue
+++ b/src/components/folio-number-input/FolioNumberInput.vue
@@ -56,9 +56,15 @@ export default class FolioNumberInput extends Vue {
     this.folioNumberString = this.folioNumber
   }
 
+  /** Update folioNumberString when prop changes  */
+  @Watch('folioNumber')
+  private onFolioNumberChange (): void {
+    this.folioNumberString = this.folioNumber
+  }
+
   /** Emit folio number on change. */
   @Watch('folioNumberString')
-  private onFolioNumberChange (): void {
+  private onFolioNumberStringChange (): void {
     this.emitFolioNumber(this.folioNumberString)
   }
 
--- a/src/components/folio-number-input/FolioNumberInput.vue
+++ b/src/components/folio-number-input/FolioNumberInput.vue
@@ -6,7 +6,7 @@
   >
     <v-text-field
       id="folio-number-textfield"
-      v-model="_folioNumber"
+      v-model="folioNumberString"
       variant="filled"
       label="Folio Number (Optional)"
       :rules="folioNumberRules"
@@ -39,7 +39,7 @@ export default class FolioNumberInput extends Vue {
   @Prop({ default: false }) readonly disabled!: boolean
 
   /** Local properties */
-  private _folioNumber: string
+  private folioNumberString: string
 
   /** Folio form model property. */
   protected folioFormValid = false
@@ -53,13 +53,13 @@ export default class FolioNumberInput extends Vue {
 
   /** Called when component is created. */
   created (): void {
-    this._folioNumber = this.folioNumber
+    this.folioNumberString = this.folioNumber
   }
 
   /** Emit folio number on change. */
-  @Watch('_folioNumber')
+  @Watch('folioNumberString')
   private onFolioNumberChange (): void {
-    this.emitFolioNumber(this._folioNumber)
+    this.emitFolioNumber(this.folioNumberString)
   }
 
   /** Prompt the field validations. */
--- a/src/components/folio-number-input/FolioNumberInput.vue
+++ b/src/components/folio-number-input/FolioNumberInput.vue
@@ -6,14 +6,13 @@
   >
     <v-text-field
       id="folio-number-textfield"
+      v-model="_folioNumber"
       variant="filled"
       label="Folio Number (Optional)"
-      :model-value="folioNumber"
       :rules="folioNumberRules"
       :disabled="disabled"
       autocomplete="chrome-off"
       :name="Math.random().toString()"
-      @update:model-value="emitFolioNumber($event)"
       @focus="emitFocus($event)"
     />
   </v-form>
@@ -39,6 +38,9 @@ export default class FolioNumberInput extends Vue {
   /** Disabled prop. */
   @Prop({ default: false }) readonly disabled!: boolean
 
+  /** Local properties */
+  private _folioNumber: string
+
   /** Folio form model property. */
   protected folioFormValid = false
 
@@ -47,6 +49,27 @@ export default class FolioNumberInput extends Vue {
     return [
       v => (!v || !this.validate || v.length <= 50) || 'Cannot exceed 50 characters' // maximum character count
     ]
+  }
+
+  /** Called when component is created. */
+  created (): void {
+    this._folioNumber = this.folioNumber
+  }
+
+  /** Emit folio number on change. */
+  @Watch('_folioNumber')
+  private onFolioNumberChange (): void {
+    this.emitFolioNumber(this._folioNumber)
+  }
+
+  /** Prompt the field validations. */
+  @Watch('folioFormValid')
+  @Watch('validate')
+  private validateField (): void {
+    if (this.validate) {
+      this.validateFolioNumber()
+      this.emitValid()
+    }
   }
 
   /** Emits an event indicating whether or not this component is valid. */
@@ -64,16 +87,6 @@ export default class FolioNumberInput extends Vue {
   @Emit('emitFolioNumber')
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   protected emitFolioNumber (val: string): void {}
-
-  /** Prompt the field validations. */
-  @Watch('folioFormValid')
-  @Watch('validate')
-  private validateField (): void {
-    if (this.validate) {
-      this.validateFolioNumber()
-      this.emitValid()
-    }
-  }
 
   /**
    * Public method that can be used through $refs from a parent
--- a/src/components/folio-number-input/package.json
+++ b/src/components/folio-number-input/package.json
@@ -5,8 +5,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@bcrs-shared-components/interfaces": "^2.0.0",
-    "vue": "^3.3.4"
+    "@bcrs-shared-components/interfaces": "^2.0.0"
   },
   "devDependencies": {
     "vue-facing-decorator": "^3.0.3"
--- a/src/components/folio-number-input/package.json
+++ b/src/components/folio-number-input/package.json
@@ -6,9 +6,9 @@
   },
   "dependencies": {
     "@bcrs-shared-components/interfaces": "^2.0.0",
-    "vue": "^2.7.14"
+    "vue": "^3.3.4"
   },
   "devDependencies": {
-    "vue-property-decorator": "^9.1.2"
+    "vue-facing-decorator": "^3.0.3"
   }
 }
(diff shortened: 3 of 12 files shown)


src/components/folio-number-input/FolioNumberInput.stories.ts
```diff
@@ -1,18 +1,21 @@
...
+    return { args }
+  },
+  template: '<folio-number-input v-bind="args" />'
 })
 
 export const Default = Template.bind({})
 Default['args'] = {
-  folioNumber: '01234567890123456789012345678901234567890123456789ABC@123/*!$ ...'
+  folioNumber: '01234567890123456789012345678901234567890123456789ABC@123/*!$ ...',
+  validate: false
```
leodube-aot, line 20: Added to show the storybook toggle for easier testing. Same for in the StaffPayment storybook file.

src/components/folio-number-input/FolioNumberInput.vue
leodube-aot, line 1: For some reason I was having issues getting the normal Vue 3 migration changes to work, mainly updating `:value` to `:model-value` and `@input` to `@update:model-value`. When I attempted those changes here (and in StaffPayment), the model would not update. Here's an example:

https://github.com/bcgov/bcrs-shared-components/assets/122323255/8a8304e6-4d84-404a-815c-71b5d0033b44

The only way I could get it to work was to use `v-model`. In this case, I had to add a `Watcher` (line 60) to emit the updated value. I also added a local variable to handle the state changes (`folioNumberString`).

leodube-aot, line 1: For some reason I was having issues getting the normal Vue 3 migration changes to work, mainly updating `:value` to `:model-value` and `@input` to `@update:model-value`. When I attempted those changes here (and in StaffPayment), the model would not update. Here's an example:

https://github.com/bcgov/bcrs-shared-components/assets/122323255/fff6dfd2-c3f0-4e38-b1c6-572ff17604df

The only way I could get it to work was to use `v-model`. In this case, I had to add a `Watcher` (line 60) to emit the updated value. I also added a local variable to handle the state changes (folioNumberString).

src/components/folio-number-input/FolioNumberInput.vue
```diff
@@ -49,6 +51,27 @@ export default class FolioNumberInput extends Vue {
...
+
+  /** Prompt the field validations. */
+  @Watch('folioFormValid')
+  @Watch('validate')
+  private validateField (): void {
+    if (this.validate) {
+      this.validateFolioNumber()
+      this.emitValid()
+    }
+  }
```
leodube-aot, line 73: Grouped the Watcher methods together.

src/components/folio-number-input/FolioNumberInput.vue
```diff
@@ -97,7 +110,7 @@ export default class FolioNumberInput extends Vue {
    * @returns True if form is valid and False if not
    */
   public validateFolioNumber (): boolean {
-    return this.$refs.folioForm.validate()
+    return this.$refs.folioForm?.validate()
```
leodube-aot, line 113: Was throwing a console error before.

src/components/staff-payment/StaffPayment.vue
```diff
@@ -18,7 +18,7 @@
         :sm="displaySideLabel ? 9 : 12"
       >
         <v-radio-group
-          v-model="paymentOption"
+          v-model="staffPaymentData.option"
```
leodube-aot, line 21: Rather than using `paymentOption`, and having to keep it updated with the `staffPaymentData` object, just use `staffPaymentData.option`

src/components/staff-payment/StaffPayment.vue
```diff
@@ -35,14 +35,13 @@
           >
             <v-text-field
               id="routing-slip-number-textfield"
-              filled
+              v-model="staffPaymentData.routingSlipNumber"
```
leodube-aot, line 38: Same issue updating `:value` and `@input` as in FolioNumberInput. Read the comment above for more info on why I move to `v-model` with a Watcher to emit the changed value.

src/components/folio-number-input/FolioNumberInput.vue
```diff
@@ -39,6 +38,9 @@ export default class FolioNumberInput extends Vue {
   /** Disabled prop. */
   @Prop({ default: false }) readonly disabled!: boolean
 
+  /** Local properties */
+  private folioNumberString: string
```
leodube-aot, line 42: Use local property to handle state changes. Otherwise we get this error:

Screenshot 2024-01-03 at 8 37 56 PM

src/components/staff-payment/StaffPayment.vue
```diff
@@ -98,22 +94,20 @@
...
-            <v-checkbox
-              id="priority-checkbox"
-              class="priority-checkbox mt-6 pt-0"
-              label="Priority (add $100.00)"
-              hide-details
-              :input-value="staffPaymentData.isPriority"
-              :disabled="paymentOption === StaffPaymentOptions.NO_FEE"
-              @change="emitStaffPaymentData({ isPriority: !!$event })"
-            />
-          </template>
```
leodube-aot, line 115: This had to be moved outside of the `v-radio-group`. For some reason it was binding to `staffPaymentData.option` and changing it's value when clicked.

src/components/staff-payment/StaffPayment.vue
```diff
@@ -214,17 +217,13 @@ export default class StaffPayment extends Vue {
         this.$refs.folioNumberInputRef.resetFolioNumberValidation()
         // enable validation for this form
         this.$refs.fasForm.validate()
-        // update data
-        this.emitStaffPaymentData({ option: StaffPaymentOptions.FAS })
```
leodube-aot, line 218: `this.emitStaffPaymentData` is now called only from `onStaffPaymentDataChanged`. This simplifies the flow a bit, as `onStaffPaymentDataChanged` will always be triggered when `onPaymentOptionChanged` changes.

src/components/staff-payment/StaffPayment.vue
leodube-aot, line 1: Quite a few changes to this file as there was some wonky stuff happening to the state. I tried to add many comments to explain why I made certain changes.

src/components/staff-payment/StaffPayment.vue
```diff
@@ -233,46 +232,27 @@ export default class StaffPayment extends Vue {
...
-
-  /** Watches for change to BCOL form validity. */
-  @Watch('bcolFormValid')
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  private onBcolFormValid (val: boolean) {
-    // ignore initial condition
-    if (!this.isMounted) return
-    this.emitValid()
-  }
-
```
leodube-aot, line 258: Moved these to be near the `valid` emit.

src/components/staff-payment/StaffPayment.vue
```diff
@@ -233,46 +232,27 @@ export default class StaffPayment extends Vue {
...
   /** Watches for changes to Staff Payment Data prop. */
   @Watch('staffPaymentData', { deep: true, immediate: true })
   private async onStaffPaymentDataChanged (val: StaffPaymentIF): Promise<void> {
-    this.paymentOption = val.option
-    await this.$nextTick()
+    this.emitStaffPaymentData({ ...this.staffPaymentData })
```
leodube-aot, line 243: Here I spread the current `staffPaymentData` for the `emitStaffPaymentData` params since = the emit function has default values to handle possible `null` values.

src/components/staff-payment/StaffPayment.vue
```diff
@@ -233,46 +232,27 @@ export default class StaffPayment extends Vue {
...
-    bcolAccountNumber = this.staffPaymentData.bcolAccountNumber || '',
-    datNumber = this.staffPaymentData.datNumber || '',
-    folioNumber = this.staffPaymentData.folioNumber || '',
-    isPriority = this.staffPaymentData.isPriority || false
+    option = StaffPaymentOptions.NONE,
+    routingSlipNumber = '',
+    bcolAccountNumber = '',
+    datNumber = '',
+    folioNumber = '',
+    isPriority = false
```
leodube-aot, line 255: The defaults on the left hand side of the `||` sign are no longer needed, since the current value from `staffPaymentData` will always get passed in line 243. We only need to handle null values.

src/components/sub-menu/SubMenu.stories.ts
leodube-aot, line 1: I decided to update SubMenu as well since it was showing an error on the StaffPayment storybook page every time I refreshed. Not sure why it was showing the error as the two components are unrelated, but it's fixed now.

Screenshot 2024-01-03 at 3 46 08 PM

src/components/sub-menu/SubMenu.vue
```diff
@@ -1,48 +1,46 @@
...
         @click="openMenu=true"
       >
         {{ name }}
       </v-btn>
     </template>
     <v-list>
-      <template v-for="(item, index) in menuItems">
+      <template
+        v-for="(item, index) in menuItems"
+        :key="index"
```
leodube-aot, line 37: `:key` should be on template tag in Vue 3, not on the children: https://v3-migration.vuejs.org/breaking-changes/key-attribute.html

src/components/folio-number-input/FolioNumberInput.vue
severinbeauvais: Yes, this happens when this code runs before the component is rendered (ie, folioForm doesn't exist yet).

This is a good solution if validate() doesn't need to be called initially...?

src/components/folio-number-input/FolioNumberInput.vue
severinbeauvais: I think we should have another set of eyes looking into this. The [Vue 3 Migration Guide](https://v3-migration.vuejs.org/breaking-changes/v-model.html) clearly indicates that v-model is changed on custom components such as this. I'm wondering if this code is actually running in Vue3?

src/components/folio-number-input/FolioNumberInput.vue
severinbeauvais: Right, because you can't set `folioNumber` because it's a read-only prop.

src/components/folio-number-input/FolioNumberInput.vue
```diff
@@ -49,6 +51,27 @@ export default class FolioNumberInput extends Vue {
     ]
   }
 
+  /** Called when component is created. */
+  created (): void {
+    this.folioNumberString = this.folioNumber
```
severinbeauvais, line 56: You should probably use another watcher on the prop, since this code runs only once (when this component is created), which assumes:
- the prop initially has the value you want
- the prop never changes

I don't think these are safe assumptions. Or at least, we should use a consistent code pattern for using props as v-models.

src/components/folio-number-input/FolioNumberInput.vue
```diff
@@ -49,6 +51,27 @@ export default class FolioNumberInput extends Vue {
...
+  }
+
+  /** Emit folio number on change. */
+  @Watch('folioNumberString')
+  private onFolioNumberChange (): void {
+    this.emitFolioNumber(this.folioNumberString)
+  }
```
severinbeauvais, line 63: I am OK with this, if we can't get `update:modelValue` working soon.

src/components/folio-number-input/package.json
```diff
@@ -6,9 +6,9 @@
   },
   "dependencies": {
     "@bcrs-shared-components/interfaces": "^2.0.0",
-    "vue": "^2.7.14"
+    "vue": "^3.3.4"
```
severinbeauvais, line 9: I think this component doesn't depend on (import) Vue any more, so this can be removed.

src/components/staff-payment/StaffPayment.vue
severinbeauvais: Pretty slick, not so much here but below where you are able to remove the `:input` event handlers 👍

src/components/staff-payment/StaffPayment.vue
```diff
@@ -168,15 +162,24 @@ export default class StaffPayment extends Vue {
...
   readonly staffPaymentData!: StaffPaymentIF
 
-  /** Radio group model property. */
-  private paymentOption = StaffPaymentOptions.NONE
-
   /** FAS form model property. */
   private fasFormValid = false
 
+  /** If FAS form inputs are disabled. */
+  private get fasFormInputDisabled (): boolean {
```
severinbeauvais, line 169: The `private` specifier may cause warnings in the template. Also, it doesn't matter if code outside this class calls a getter. So I recommend leaving computed variables as public.

src/components/staff-payment/StaffPayment.vue
```diff
@@ -168,15 +162,24 @@ export default class StaffPayment extends Vue {
...
+    return this.staffPaymentData.option === StaffPaymentOptions.BCOL ||
+      this.staffPaymentData.option === StaffPaymentOptions.NO_FEE
+  }
+
   /** BCOL form model property. */
   private bcolFormValid = false
```
severinbeauvais, line 175: I recommend keeping local variables together, and computeds (getters) together (and near the top of the class).

src/components/staff-payment/StaffPayment.vue
```diff
@@ -287,6 +267,24 @@ export default class StaffPayment extends Vue {
     }
   }
 
+  /** Watches for change to FAS form validity. */
+  @Watch('fasFormValid')
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
severinbeauvais, line 272: Instead of this, could you just have no formal parameters on line 273 (and 282)?

src/components/staff-payment/package.json
```diff
@@ -8,9 +8,9 @@
     "@bcrs-shared-components/enums": "^2.0.0",
     "@bcrs-shared-components/folio-number-input": "^2.0.0",
     "@bcrs-shared-components/interfaces": "^2.0.0",
-    "vue": "^2.7.14"
+    "vue": "^3.3.4"
```
severinbeauvais, line 11: As above, I think this isn't needed, as it's not imported by the component.

src/components/sub-menu/SubMenu.vue
```diff
@@ -1,48 +1,46 @@
 <template>
   <v-menu
     :close-on-content-click="false"
-    :offset-x="isOffsetX"
-    :offset-y="isOffsetY"
     :open-on-hover="isOpenOnHover"
     :transition="transition"
-    :value="openMenu"
+    :model-value="openMenu"
```
severinbeauvais, line 6: So this works and it's the update event that doesn't work?

src/components/sub-menu/package.json
```diff
@@ -5,9 +5,9 @@
     "access": "public"
   },
   "dependencies": {
-    "vue": "^2.7.14"
+    "vue": "^3.3.4"
```
severinbeauvais, line 8: Not imported.

severinbeauvais (commented):
Great work! I'm concerned about the v-model thing and would like someone else to also look at it (though your work-around looks like it works).

src/components/sub-menu/SubMenu.stories.ts
ketaki-deodhar: Thanks for fixing this :)

src/components/folio-number-input/FolioNumberInput.vue
ketaki-deodhar: Yes. Looks like workaround works okay. I can take a look parallelly :)

severinbeauvais: Thanks! 🙏 ❤️

ketaki-deodhar:
Please add `color="primary"` to `v-radio` so that the color is changed to blue after selection. I had the same issue in LimitedRestorationPanel upgrade :)

DEV:
image

LOCAL:
image

src/components/folio-number-input/FolioNumberInput.vue
ketaki-deodhar: @leodube-aot , can you try this :
image

leodube-aot: @ketaki-deodhar My initial testing with this seems to work! I will try to update the other components in this PR to see if the change will work for them as well.

src/components/folio-number-input/FolioNumberInput.vue
leodube-aot: I removed the `folioNumberString` watcher since we got the `update:folioNumberString` working.

src/components/staff-payment/StaffPayment.vue
leodube-aot: For this one, I think it is actually better to keep it as v-model. This is because the original code used the `update` function to emit the staff payment data. This is now done in the watcher on `staffPaymentData`, which cleans up the code a bit since the emit function is called from only one place.

src/components/staff-payment/StaffPayment.vue
leodube-aot: I removed the parameters!

src/components/sub-menu/SubMenu.vue
leodube-aot: Yes it appears that the update event was mainly causing the issues. Ketaki's solution above seems to be the fix.

src/components/folio-number-input/package.json
leodube-aot: Removed!

src/components/folio-number-input/FolioNumberInput.vue
severinbeauvais: That DatePicker looks a bit different because the prop is explicitly called "initial value".

Generally, yes, copying props to local variables in created() or mounted() isn't enough; you also need to handle props changing (ie, reactivity).

src/components/folio-number-input/FolioNumberInput.vue
ketaki-deodhar: Awesome! Sounds good Leo

ketaki-deodhar (approved):
Looks good!

severinbeauvais:
Please rebase.

Is it all working properly now?

JazzarKarim (approved):
LGTM!

leodube-aot:
@severinbeauvais Yes all 3 components seem to be working fine now

severinbeauvais (approved):
Good job. Anything lessons learned for other devs?

leodube-aot:
I think the main takeaway here is that during the conversion of the `v-model` prop and event default names from:
- `:value="someVariable"` -> `:modelValue="someVariable"`
- `@input="someFunction($event)"` -> `@update:modelValue="someFunction($event)`

if the updates to the `modelValue` prop are not working correctly, then updating the event to:
- `@input="someFunction($event)"` -> `@update:someVariable="someFunction($event)`

is the solution.

PR #232, opened 2024-01-08
18997: Vue 3 Upgrade - Completing Party

PaulGarewal:
*Issue #:* /bcgov/entity#18997

*Description of changes:*

- Upgrade the Completing Party Shared Component to Vue3
- Used the Base-Address Vue 3 component upgrade as a template; excessive imports here that may need to be cut to reduce the bloat.


PaulGarewal:
Current output in Storybook:

image

From the Vue 2 version (below) I am unable to currently display the address:

image

May be due to the excessive Address files and importing the AddressIF incorrectly. Currently working on this and reducing some bloat in the supporting files.

src/components/completing-party/factories/address-factory.ts
severinbeauvais, line 1: I think these files shouldn't be here. They seem to be a duplicate of the files in base-address/factories.

The completing party component should be able to import everything it needs from the sub-components it uses (such as Base Address). The completing party component should know the minimum about the sub-components it uses -- only its interface (props and events).

Ditto for the resources folder below.

PaulGarewal: Removed and storybook runs no problem, currently tinkering with mailing address

src/components/completing-party/CompletingParty.stories.ts
```diff
@@ -1,43 +1,19 @@
-import type { Meta } from '@storybook/vue'
-import { CompletingParty } from './index'
-import { required, maxLength } from 'vuelidate/lib/validators'
+import { Meta } from '@storybook/vue3'
+import CompletingParty from './CompletingParty.vue'
+import { AddressIF } from '@bcrs-shared-components/interfaces'
+import { DefaultSchema } from '../base-address/resources/default-schema'
```
severinbeauvais, line 4: Should a normal component (in one of our UIs) be able to import the default schema? If so then it should be exported as part of Base Address. Or, if it's just used here for testing purposes then this is OK.

JazzarKarim (approved):
LGTM

src/components/base-address/BaseAddress.vue
```diff
@@ -169,8 +169,9 @@ import {
   useBaseValidations,
   spaceRules
 } from '@/components/base-address/factories'
-import { AddressIF, SchemaIF } from '@bcrs-shared-components/interfaces'
-import { AddressValidationRules } from '@bcrs-shared-components/enums'
+import { AddressIF } from '@bcrs-shared-components/interfaces'
+import { SchemaIF } from '@/interfaces'
```
severinbeauvais, line 173: Can you import this from the shared folder?

In Storybook (ie, in this repo), this import will work because `@/interfaces` resolves to "./src/interfaces" (ie, local folder) [link to config](https://github.com/bcgov/bcrs-shared-components/blob/62a05e2e40024ec2cf6095285332f5811912071a/vite.config.ts#L19).

However, when this component is imported into another UI, `@/interfaces` will be that UI's local folder, which probably doesn't contain SchemaIF.

src/components/base-address/factories/address-factory.ts
```diff
@@ -187,21 +187,22 @@ export function useCountryRegions (code: string): boolean {
 }
 
 export function formatAddress (address: AddressIF): AddressIF {
-  address.postalCode = address.postalCode?.toUpperCase() || ''
-  if (address.country === 'CA') {
+  let formattedPostalCode = address.postalCode?.toUpperCase() || ''
```
severinbeauvais, line 190: Yes, don't change the passed-in address object. Good 👍

src/components/base-address/factories/address-factory.ts
```diff
@@ -187,21 +187,22 @@ export function useCountryRegions (code: string): boolean {
...
-  return {
-    country: address.country?.trim(),
-    street: address.street?.trim(),
-    streetAdditional: address.streetAdditional?.trim(),
-    city: address.city?.trim(),
-    region: address.region?.trim(),
+  return{
+    addressCity: address.addressCity?.trim(),
+    addressCountry: address.addressCountry?.trim(),
+    addressRegion: address.region?.trim(), // FUTURE: change to undefined?
```
severinbeauvais, line 200: Should this be `address.addressRegion?.trim()` ?

Why would you set this to undefined? (It will automatically be undefined if the optional chaining fails.)

src/components/base-address/factories/address-factory.ts
```diff
@@ -187,21 +187,22 @@ export function useCountryRegions (code: string): boolean {
...
+    addressType: null, // FUTURE: change to undefined?
```
severinbeauvais, line 201: Why not pass this through?
`addressType: address.addressType`

(Theoretically this is an enum, but I see in the interface that it's declared as a string 😞 )

severinbeauvais (commented):
OK, I see there was a mismatch between address objects. How far did this get you -- are both BaseAddress and CompletingParty components working now?

src/components/base-address/BaseAddress.vue
PaulGarewal: I just changed that because it was throwing an error in hopes to solve the problem, I will change it back.

src/components/base-address/factories/address-factory.ts
PaulGarewal: I believe the undefined is carry over from Assets Team, I will change these.

src/components/base-address/BaseAddress.vue
```diff
@@ -244,15 +245,15 @@ export default defineComponent({
         localSchema.postalCode = origPostalCodeRules.concat([baseRules.zipCode])
         localSchema.region = origRegionRules
       } else {
-        localSchema.postalCode = origPostalCodeRules.concat([baseRules[AddressValidationRules.MAX_LENGTH](15)])
-        localSchema.region = [baseRules[AddressValidationRules.MAX_LENGTH](2), ...spaceRules]
+        localSchema.postalCode = origPostalCodeRules.concat([baseRules[AddressValidationRules.MAX_LENGTH]])
+        localSchema.region = [baseRules[AddressValidationRules.MAX_LENGTH], ...spaceRules]
```
severinbeauvais, line 249: Old code looked weird. Does new code work as expected?

src/components/base-address/factories/address-factory.ts
severinbeauvais: Wait. Should you be using `formattedPostalCode` on line 203?

The other thing you could do is move toUpperCase() into line 203 (chaining).

src/components/completing-party/CompletingParty.stories.ts
severinbeauvais: Obsolete code. Resolving.

src/components/base-address/BaseAddress.vue
severinbeauvais: Reverted. Thanks. Resolving.

src/components/base-address/BaseAddress.vue
```diff
@@ -170,7 +170,7 @@ import {
   spaceRules
 } from '@/components/base-address/factories'
 import { AddressIF, SchemaIF } from '@bcrs-shared-components/interfaces'
-import { AddressValidationRules } from '@bcrs-shared-components/enums'
+import { AddressValidationRules } from '@bcrs-shared-components/enums/address-validation-rules'
```
severinbeauvais, line 173: Why do you need to make this change? Looks like old code should work...?

src/components/base-address/factories/address-factory.ts
severinbeauvais: Looks good. Resolving.

severinbeauvais:
Re:
image

If you can't resolve this with a rebase, just revert the files and remind me to do a `lerna bootstrap --hoist` when I next publish these components.

src/components/base-address/BaseAddress.vue
PaulGarewal: Gave me an error message before, now with the change it doesn't

severinbeauvais: Weird, because AddressValidationRules is exported correctly via the index.ts file.

What's the error message?

PaulGarewal: "has no exported member named x", which is double weird because I could import ApprovalTypes no problem when trying. SchemaIF also gets the same error message, while AddressIF does not even though they are in the same file. Might just be the editor but I am unsure.

severinbeauvais: I see that the BaseAddress component does not import the shared enums. Try adding that (plus `lerna bootstrap --hoist`).

severinbeauvais: Sorry, I was looking at wrong branch.

Try that Lerna command anyway,

src/components/base-address/BaseAddress.vue
```diff
@@ -170,7 +170,7 @@ import {
   spaceRules
 } from '@/components/base-address/factories'
```
severinbeauvais, line 171: This might be a problem. The path resolves to "src/components/base-address/factories", which is fine in this repo (ie, for Storybook's usage) but won't work when Base Address is imported in another UI.

Try changing this to `from './factories`.

src/components/base-address/BaseAddress.vue
severinbeauvais: The new code gives an error in my VS Code. The old code seems to be fine.

PaulGarewal: Weird, I have the vice versa 👀

src/components/base-address/BaseAddress.vue
PaulGarewal: That worked great, thanks 👍

src/components/base-address/BaseAddress.vue
severinbeauvais: By the way, the old code works fine for me, so you shouldn't need to do this either.

src/components/base-address/BaseAddress.vue
severinbeauvais: Great. Resolving.

src/components/base-address/BaseAddress.vue
severinbeauvais: The `(15)` and `(2)` are parameters to the function.

In your editor, if you hover on `baseRules`, does it appear as an object of functions?

severinbeauvais:
By the way, `lerna bootstrap --hoist` is deprecated in the latest version of Lerna. One of us will have to research what the new way to update component dependencies is.

src/components/base-address/BaseAddress.stories.ts
```diff
@@ -18,26 +18,30 @@ const Template = (args: AddressIF) => ({
...
 export const WithData = Template.bind({})
 WithData['args'] = {
-  street: '1234 Sesame Street',
-  streetAdditional: '4th Floor',
-  city: 'Victoria',
-  region: 'BC',
-  postalCode: 'V8N 1A1',
-  country: 'Canada',
-  deliveryInstructions: 'Leave at front door',
+  value: {
```
JazzarKarim, line 36: The story for BaseAddress was wrong since the props were incorrect. The new one from assets uses "value". 

Before:
before with data

Now:
after with data

src/components/completing-party/CompletingParty.vue
```diff
@@ -69,7 +69,7 @@
                 ref="mailingAddress"
                 :editing="true"
                 :schema="addressSchema"
-                :address="completingParty.mailingAddress"
+                :value="completingParty.mailingAddress"
```
JazzarKarim, line 72: Same as above. We use value now,

src/components/completing-party/CompletingParty.stories.ts
```diff
@@ -31,5 +32,8 @@ Default['args'] = {
       streetAddressAdditional: 'Test Street Additional'
     }
   },
-  editing: true
+  enableAddEdit: false,
+  addressSchema: DefaultSchema,
```
JazzarKarim, line 36: Looks good now:

https://github.com/bcgov/bcrs-shared-components/assets/122301442/e9c639f2-dbf3-4cd2-bc2d-f6240ba7b070

severinbeauvais:
Karim, please verify and close off the other conversations. Nice work so far!

src/components/base-address/factories/address-factory.ts
JazzarKarim: This should be fixed. Using formattedPostalCode now on line 203:
https://github.com/bcgov/bcrs-shared-components/blob/4a37dd52699fa636fe23ba61beb0413179c74daf/src/components/base-address/factories/address-factory.ts#L203

Closing the conversation.

src/components/base-address/factories/address-factory.ts
JazzarKarim: Now it's `address.addressRegion?.trim()`:
https://github.com/bcgov/bcrs-shared-components/blob/4a37dd52699fa636fe23ba61beb0413179c74daf/src/components/base-address/factories/address-factory.ts#L200

Resolving.

src/components/base-address/BaseAddress.vue
JazzarKarim: Old code works. I reverted this. Resolving.

src/components/base-address/BaseAddress.vue
JazzarKarim: This was reverted back so I think it's OK. Resolving this.

JazzarKarim:
> Karim, please verify and close off the other conversations. Nice work so far!

Thanks! I verified all conversations. Please let me know if there are any other comments or if I can proceed with merging.

severinbeauvais (approved):
Great work! Merge when ready.

src/components/completing-party/CompletingParty.stories.ts
PaulGarewal: Well done Karim! You're a hero ❤️

PR #245, opened 2024-01-29
Fix base address, so it emits the update-address event. Convert to sc…

seeker25:
…ript setup. Get rid of Vue2 dep in interfaces. Interfaces shouldn't reference Vue.


src/components/base-address/BaseAddress.vue
```diff
@@ -174,153 +174,145 @@
...
 
-    const { enableAddressComplete, uniqueIds } = useAddressComplete(addressLocal)
+// eslint-disable-next-line vue/no-setup-props-destructure
+const localSchema = { ...props.schema }
+const {
+  addressLocal,
+  country,
+  schemaLocal,
+  labels
+} = useAddress(toRef(props.value) as Ref<AddressIF>, localSchema)
```
JazzarKarim, line 232: Nice!

src/interfaces/package.json
```diff
@@ -5,7 +5,6 @@
     "access": "public"
   },
   "dependencies": {
-    "@bcrs-shared-components/enums": "^2.0.0",
-    "vue": "^2.7.14"
```
JazzarKarim, line 9: 👍

PR #253, opened 2024-05-07
20989 - Add USA states to jurisdiction list

ketaki-deodhar:

*Description of changes:*

- Add USA states to jurisdiction list
- Add a flag to show USA jurisdictions conditionally

Flag defaults to flase (Current implementation for NR and Amalgamation)

image

Flag set to True (lists USA states)

image

image

### DIFF
--- a/src/components/jurisdiction/Jurisdiction.vue
+++ b/src/components/jurisdiction/Jurisdiction.vue
@@ -13,7 +13,7 @@
 import { Component, Emit, Prop, Vue } from 'vue-property-decorator'
 import NestedSelect from './NestedSelect.vue'
 import { JurisdictionLocation } from '@bcrs-shared-components/enums'
-import { CanJurisdictions, IntlJurisdictions } from './list-data'
+import { CanJurisdictions, IntlJurisdictions, UsaJurisdiction } from './list-data'
 
 @Component({
   components: { NestedSelect }
@@ -22,6 +22,7 @@ export default class Jurisdiction extends Vue {
   // props
   @Prop({ default: 'Select the home jurisdiction' }) readonly label!: string
   @Prop() readonly errorMessages!: string
+  @Prop({ default: false }) readonly showUsJurisdictions!: boolean
 
   // variables
   jurisdiction = null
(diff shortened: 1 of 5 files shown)


src/components/jurisdiction/Jurisdiction.vue
```diff
@@ -41,15 +42,26 @@ export default class Jurisdiction extends Vue {
...
+    array.push({ isHeader: true, group: 2, text: 'International' })
     IntlJurisdictions
       .filter(jur => jur.value !== JurisdictionLocation.CA)
       .forEach(jur => array.push({
-        group: 1,
+        group: 2,
         text: jur.text,
         value: jur.value,
-        separator: false
+        separator: (jur.value === JurisdictionLocation.IN)
```
ketaki-deodhar, line 64: Added this to show a separator between for International

src/components/jurisdiction/Jurisdiction.vue
```diff
@@ -41,15 +42,26 @@ export default class Jurisdiction extends Vue {
...
+      array.push({ isHeader: true, group: 1, text: 'United States' })
+      UsaJurisdiction
+        .forEach(jur => array.push({
+          group: 1,
+          text: jur.text,
+          value: jur.value,
+          separator: false
+        }))
+    }
+
```
ketaki-deodhar, line 55: Yui confirmed to about Puerto Rico and VI: "Please use the same as live Name Request. I.e. Puerto Rico and VI are considered International."

src/components/jurisdiction/Jurisdiction.vue
```diff
@@ -41,15 +42,26 @@ export default class Jurisdiction extends Vue {
...
     // add in International jurisdictions (not including CA)
-    array.push({ isHeader: true, group: 1, text: 'International' })
+    array.push({ isHeader: true, group: 2, text: 'International' })
     IntlJurisdictions
       .filter(jur => jur.value !== JurisdictionLocation.CA)
```
severinbeauvais, line 59: Did you find out why you don't need to filter out USA when `showUsJurisdictions` is True?

ketaki-deodhar: No not yet. Just wondering if somehow it is removed in `filter` or `forEach` function considering duplicates in the Array. (one added when adding USA jurisdiction and other in International jurisdiction list)

JazzarKarim (approved):
LGTM 👍

src/components/jurisdiction/Jurisdiction.vue
severinbeauvais: Those filter and forEach calls are separate, and working on separate lists. I dunno. Might look at it this afternoon for fun.

severinbeauvais (approved):
I approve since this is working correctly.

I will satisfy myself of _why_ it's working after this is merged and published.

severinbeauvais:
While publishing, I bumped up the minor version since this is more than a patch.

image

ketaki-deodhar:
> While publishing, I bumped up the minor version since this is more than a patch.
> 
> image

Thanks Sev!

severinbeauvais:
So... US is in group 2 even when the prop is True.

image

severinbeauvais:
@ketaki-deodhar , you were on the right track with duplicates -- the list doesn't support duplicate text items, but "United States" was in there first as a group and then as an item.

I consider this a bug. It can be demonstrated by changing the text (label) of the group, ie,

image

Please add a filter to remove US from the international list when the prop is True. (For cleanliness, I suggest creating a nested function and passing it to the filter call.)

PR #254, opened 2024-05-07
20989 - Exclude USA conditionally

ketaki-deodhar:

*Description of changes:*

- new function to exclude USA conditionally
- pass newly created function to filter call

**Changed header text to 'USA' for Testing only**. Does not have United Sates in the International jurisdiction list anymore:

image

### DIFF
--- a/src/components/jurisdiction/Jurisdiction.vue
+++ b/src/components/jurisdiction/Jurisdiction.vue
@@ -42,6 +42,7 @@ export default class Jurisdiction extends Vue {
         separator: (jur.value === JurisdictionLocation.FD)
       }))
 
+    // add USA jurisdictions (conditionally)
     if (this.showUsJurisdictions) {
       array.push({ isHeader: true, group: 1, text: 'United States' })
       UsaJurisdiction
@@ -56,7 +57,7 @@ export default class Jurisdiction extends Vue {
     // add in International jurisdictions (not including CA)
     array.push({ isHeader: true, group: 2, text: 'International' })
     IntlJurisdictions
-      .filter(jur => jur.value !== JurisdictionLocation.CA)
+      .filter(jur => !this.excludeJurisdictions(jur.value).includes(jur.value))
       .forEach(jur => array.push({
         group: 2,
         text: jur.text,
@@ -65,6 +66,23 @@ export default class Jurisdiction extends Vue {
       }))
 
     return array
+  }
+
+  /**
+   * Always exclude CA
+   * Exclude USA when states are listed in the jurisdiction list based on showUsJurisdictions flag
+   */
+  excludeJurisdictions (jurisdiction): Array<any> {
+    const excludedValues = []
+    if (jurisdiction === JurisdictionLocation.CA) {
+      excludedValues.push(JurisdictionLocation.CA)
+    }
+
+    if (this.showUsJurisdictions) {
+      excludedValues.push(JurisdictionLocation.US)
+    }
+
+    return excludedValues
   }
 
   @Emit('change')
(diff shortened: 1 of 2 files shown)


ketaki-deodhar:
@severinbeauvais thanks for spotting this bug.

src/components/jurisdiction/Jurisdiction.vue
```diff
@@ -56,7 +57,7 @@ export default class Jurisdiction extends Vue {
     // add in International jurisdictions (not including CA)
     array.push({ isHeader: true, group: 2, text: 'International' })
     IntlJurisdictions
-      .filter(jur => jur.value !== JurisdictionLocation.CA)
+      .filter(jur => !this.excludeJurisdictions(jur.value).includes(jur.value))
```
severinbeauvais, line 60: I don't understand why you exclude and then include.

src/components/jurisdiction/Jurisdiction.vue
```diff
@@ -67,6 +68,23 @@ export default class Jurisdiction extends Vue {
...
+   * Always exclude CA
+   * Exclude USA when states are listed in the jurisdiction list based on showUsJurisdictions flag
+   */
+  excludeJurisdictions (jurisdiction): Array<any> {
+    const excludedValues = []
+    if (jurisdiction === JurisdictionLocation.CA) {
+      excludedValues.push(JurisdictionLocation.CA)
+    }
+
+    if (this.showUsJurisdictions) {
```
JazzarKarim, line 81: I wonder if we can rename this to be more explicit. When I first read it, it sounded to me "show us jurisdictions" as in show the reader jurisdictions 😆

JazzarKarim (approved):
👍 👍

src/components/jurisdiction/Jurisdiction.vue
```diff
@@ -67,6 +68,23 @@ export default class Jurisdiction extends Vue {
...
+      excludedValues.push(JurisdictionLocation.US)
+    }
+
+    return excludedValues
+  }
```
severinbeauvais, line 86: Whoa, this gets called for every item in the Intl list. And it does a lot of work.

Try something simpler.

src/components/jurisdiction/Jurisdiction.vue
severinbeauvais: Usa?

JazzarKarim: Yep, was thinking that too. That would be great.

src/components/jurisdiction/Jurisdiction.vue
severinbeauvais: I think you only need a simple conditional here that returns False if `filter` should exclude it, else True.

src/components/jurisdiction/Jurisdiction.vue
severinbeauvais: Code removed. Resolving.

src/components/jurisdiction/Jurisdiction.vue
severinbeauvais: LGTM

src/components/jurisdiction/Jurisdiction.vue
severinbeauvais: Looks good.

severinbeauvais (approved):

severinbeauvais:
image

src/components/jurisdiction/Jurisdiction.vue
JazzarKarim: Thanks for fixing this.

PR #255, opened 2024-05-09
20989-add initial value prop

ketaki-deodhar:

*Description of changes:*

Add initial value prop to retrieve value for a draft filing.

some initial value tests:

image

image

image

image

### DIFF
--- a/src/components/jurisdiction/Jurisdiction.stories.ts
+++ b/src/components/jurisdiction/Jurisdiction.stories.ts
@@ -16,3 +16,14 @@ export const Default = Template.bind({})
 Default['args'] = {
   showUsaJurisdictions: false
 }
+
+export const JurisdictionInitialValue = Template.bind({})
+JurisdictionInitialValue['args'] = {
+  showUsaJurisdictions: false,
+  initialValue: {
+    group: 0,
+    text: 'Alberta',
+    value: 'AB',
+    separator: false
+  }
+}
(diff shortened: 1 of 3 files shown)


src/components/jurisdiction/Jurisdiction.stories.ts
```diff
@@ -16,3 +16,14 @@ export const Default = Template.bind({})
 Default['args'] = {
   showUsaJurisdictions: false
 }
+
+export const JurisdictionInitialValue = Template.bind({})
+JurisdictionInitialValue['args'] = {
+  showUsaJurisdictions: false,
+  initialValue: {
```
JazzarKarim, line 23: Wait, so I see below the type of this prop a string while here it's an object?

src/components/jurisdiction/Jurisdiction.stories.ts
```diff
@@ -16,3 +16,14 @@ export const Default = Template.bind({})
...
+    group: 0,
+    text: 'Alberta',
+    value: 'AB',
+    separator: false
+  }
```
severinbeauvais, line 28: There's information in here that you won't be saving to or restoring from a draft.

I think you need to convert the initial value from `{ country, region }` (which is what the parent component provides) to { group, value }` (which is what this component understands).

src/components/jurisdiction/Jurisdiction.stories.ts
ketaki-deodhar: looking into it....

src/components/jurisdiction/Jurisdiction.stories.ts
ketaki-deodhar: Yes. You are right. Looking into it.

severinbeauvais (approved):
Looks as expected. Does it work?

ketaki-deodhar:
> Looks as expected. Does it work?

Tested with some scenario, screenshots above.

severinbeauvais:
> > Looks as expected. Does it work?
> 
> Tested with some scenarios, screenshots above.

Looks good!

PR #257, opened 2024-09-10
16079 Many fixes to Approval Type component

severinbeauvais:
*Issue #:* bcgov/entity#16079

*Description of changes:*
- added invalid-section styling
- fixed whitespace
- simplified v-radio template
- fixed rules
- fixed change/input/error handling
- added validate prop
- updated court order number rules text
- fixed initial state (when mounted)
- fixed validations
- fixed lint issues
- updated/fixed unit tests

### DIFF
--- a/src/components/approval-type/ApprovalType.vue
+++ b/src/components/approval-type/ApprovalType.vue
@@ -18,7 +18,7 @@
         class="mb-n5"
       >
         <v-radio-group
-          class="payment-group mt-0"
+          class="mt-0"
           :value="approvalTypeSelected"
           @change="setApprovalTypeSelected($event)"
         >
@@ -43,7 +43,7 @@
                 id="court-order-number-input"
                 class="mt-4"
                 label="Court Order Number"
-                :value="courtOrderNumber"
+                :value="courtOrderNumberText"
                 :rules="validate ? courtOrderNumRules : []"
                 hide-details="auto"
                 filled
@@ -71,7 +71,7 @@
                   class="mt-2"
                   title="Select Date"
                   :nudgeRight="150"
-                  :initialValue="noticeDate"
+                  :initialValue="noticeDateText"
                   :inputRules="validate ? datePickerRules : []"
                   @emitDateSync="setNoticeDateText($event)"
                 />
@@ -83,7 +83,7 @@
                   class="mt-2"
                   title="Select Date"
                   :nudgeRight="150"
-                  :initialValue="applicationDate"
+                  :initialValue="applicationDateText"
                   :inputRules="validate ? datePickerRules : []"
                   @emitDateSync="setApplicationDateText($event)"
                 />
@@ -114,7 +114,10 @@ export default class ApprovalType extends Vue {
     courtNumRef: FormIF
   }
 
-  /** The initial court order number. */
+  /**
+   * The initial court order number.
+   * Is only read when the component is mounted.
+   */
   @Prop({ default: '' }) readonly courtOrderNumber!: string
 
   /** Whether restoration was approved by the registrar. */
@@ -126,10 +129,16 @@ export default class ApprovalType extends Vue {
   /** Filing name used in radio options. */
   @Prop({ default: 'restoration' }) readonly filingType!: string
 
-  /** The initial date the notice of the application for restoration was published in the BC Gazette. */
+  /**
+   * The initial date the notice of the application for restoration was published in the BC Gazette.
+   * Is only read when the component is mounted.
+   */
   @Prop({ default: '' }) readonly noticeDate!: string
 
-  /** The initial date the application for restoration was mailed to the company. */
+  /**
+   * The initial date the application for restoration was mailed to the company.
+   * Is only read when the component is mounted.
+   */
   @Prop({ default: '' }) readonly applicationDate!: string
 
   /** Whether this section is invalid. */
@@ -194,7 +203,7 @@ export default class ApprovalType extends Vue {
   private validateComponent (): boolean {
     let valid = true
     if (this.approvalTypeSelected === ApprovalTypes.VIA_COURT_ORDER) {
-      valid = this.$refs.courtNumRef.validate()
+      valid = this.$refs.courtNumRef?.validate()
     } else if (this.approvalTypeSelected === ApprovalTypes.VIA_REGISTRAR) {
       // Valid if validation flag is not set.
       // Valid if both dates are selected.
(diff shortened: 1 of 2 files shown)


severinbeauvais:
Hello. Please review these changes while I work on the unit tests.

I am aware that some of these changes will break existing users of this component. For this reason, I will update the minor version number of this component (ie, from 1.0.61 to 1.1.0). (I can't use 2.x.x because it's already used for Vue3 code.)

src/components/approval-type/ApprovalType.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div id="approval-type">
+  <div
+    id="approval-type"
+    :class="{ 'invalid-section': invalidSection }"
+  >
```
severinbeauvais, line 5: Example:

image

src/components/approval-type/ApprovalType.vue
```diff
@@ -8,87 +11,81 @@
...
       </v-col>
+
       <v-col
         cols="12"
         sm="9"
-        class="mt-n4"
+        class="mb-n5"
       >
         <v-radio-group
-          v-model="approvalTypeSelected"
```
severinbeauvais, line 17: It seemed redundant to use both a v-model property and handle the change event so I went with `:value` and `@change` to keep things consistent with the other components in here.

src/components/approval-type/ApprovalType.vue
```diff
@@ -8,87 +11,81 @@
...
+            <v-expand-transition>
               <v-text-field
                 v-if="approvalTypeSelected === ApprovalTypes.VIA_COURT_ORDER"
                 id="court-order-number-input"
-                v-model="courtOrderNumberText"
+                class="mt-4"
                 label="Court Order Number"
-                :rules="courtOrderNumRules"
+                :value="courtOrderNumber"
+                :rules="validate ? courtOrderNumRules : []"
```
severinbeauvais, line 47: The rules are conditional on the validate prop (but you have to wait a tick for validation to happen).

src/components/approval-type/ApprovalType.vue
```diff
@@ -8,87 +11,81 @@
...
                 hide-details="auto"
                 filled
-                @input="courtOrderNumberChanged"
-                @update:error="emitValidationError($event)"
```
severinbeauvais, line 49: Redundant. Now handled more elegantly below.

src/components/approval-type/ApprovalType.vue
```diff
@@ -117,148 +114,147 @@ export default class ApprovalType extends Vue {
...
+  /** Whether restoration was approved by the registrar. */
   @Prop({ default: false }) readonly approvedByRegistrar!: boolean
 
-  /** filing name used in radio options. */
-  @Prop({ default: 'restoration' }) readonly filingType!: string
+  /** Whether restoration was approved by court order. */
+  @Prop({ default: false }) readonly approvedByCourtOrder!: boolean
 
-  /** Show only the court order option; remove via registrar option. */
-  @Prop({ default: false }) readonly isCourtOrderOnly!: boolean
```
severinbeauvais, line 130: I replaced this with `approvedByCourtOrder` (similar to approvedByRegistrar), which keeps the flag `isCourtOrderRadio`, below, functionally separate -- they do separate things instead of being dependent on each other.

src/components/approval-type/ApprovalType.vue
```diff
@@ -117,148 +114,147 @@ export default class ApprovalType extends Vue {
...
+      this.setNoticeDateText(this.noticeDate)
+      this.setApplicationDateText(this.applicationDate)
+    } else if (this.approvedByCourtOrder) {
+      // first set approval type, then set values
+      // otherwise, validation will fail
+      this.setApprovalTypeSelected(ApprovalTypes.VIA_COURT_ORDER)
+      this.setCourtOrderNumberText(this.courtOrderNumber)
     }
-    if (!this.isCourtOrderRadio) {
-      this.approvalTypeSelected = ApprovalTypes.VIA_COURT_ORDER
```
severinbeauvais, line 180: There is no more need for "default" state handling. When we come in to this component, we should have either `approvedByRegistrar` or `approvedByCourtOrder` set.

If both are set (in error) then only the first one will have an effect.

If neither are set (in error) then no radio buttons are pre-selected, but the component should still function correctly once the user selects one.

src/components/approval-type/ApprovalType.vue
```diff
@@ -117,148 +114,147 @@ export default class ApprovalType extends Vue {
...
 
-  /** Triggers the form validation. */
-  public validate (): boolean {
+  /** Validates the component and emits the validity. */
+  @Emit('valid')
+  private validateComponent (): boolean {
+    let valid = true
     if (this.approvalTypeSelected === ApprovalTypes.VIA_COURT_ORDER) {
-      let status = this.$refs.courtNumRef.validate()
-      this.$emit('valid', status)
```
severinbeauvais, line 188: We shouldn't use this all over the code. We should have a single place where a property is emitted. This is now THIS method (as per `@Emit` decorator).

severinbeauvais:
See https://github.com/bcgov/entity/issues/16079 for more screenshots.

src/components/approval-type/ApprovalType.vue
```diff
@@ -117,148 +114,147 @@ export default class ApprovalType extends Vue {
...
 
   /** Whether this section is invalid. */
   @Prop({ default: false }) readonly invalidSection!: boolean
 
-  /** Whether the display of court order section is a radio button. */
+  /** Whether to show the court order radio button. */
   @Prop({ default: true }) readonly isCourtOrderRadio!: boolean
 
+  /** Whether validation is enabled. */
+  @Prop({ default: false }) readonly validate!: boolean
```
JazzarKarim, line 142: Question about this prop. In my head, I'm thinking we never want it to be false. Why would we want to use the component with no validation enabled?

JazzarKarim (approved):
Aside from my one question, looks great!

I bet this grew in size to what was expected. Needed lots of work.

src/components/approval-type/ApprovalType.vue
severinbeauvais: In its initial state, we don't want to show validation errors. Once the user does something -- in this case, try to go to the Review and Confirm page -- then we check for and show errors.

So, with this prop False, this component is always valid. With this prop True then it depends what data has been entered (ie, the rules are applied).

JazzarKarim: Ahhh, that's brilliant. OK, makes sense, thanks! Resolving.

ArwenQin (approved):

ketaki-deodhar (approved):

PR #265, opened 2025-01-30
25408 - DocumentId Component: Include BCRS Shared Rebuild

cameron-eyds:
*Issue #:* /bcgov/entity#25408

*Description of changes:*
- Creates Shared Document Id component. 
- Includes Storybook example and unit testing.
- Minor vite configuration updates.
- Storybook Rebuild.

*Notes:* 
- Fixed misc unit testing issues

Screenshot 2025-01-30 at 10 19 32 AM


src/components/document-id/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@bcrs-shared-components/document-id",
+  "version": "0.0.0",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "axios": "^0.27.2",
+    "@bcrs-shared-components/interfaces": "^1.1.23",
```
severinbeauvais, line 9: The new component doesn't depend on interfaces, but it does depend on enums. Small update, please.

image

cameron-eyds: Gracias!

severinbeauvais (approved):
Amazing! Thanks so  much for fixing the unit tests and updating Storybook.

~I'm curious why so many awaits are now needed for some tests, as I think they weren't needed before, but it's just unit tests so I'm not going to lose any sleep over it. Thanks again!~ _Those were there before. The code has been cleaned up a bit, that's all._

cameron-eyds:
> Amazing! Thanks so much for fixing the unit tests and updating Storybook.
> 
> I'm curious why so many awaits are now needed for some tests, as I think they weren't needed before, but it's just unit tests so I'm not going to lose any sleep over it. Thanks again!

Of course, no prob at all! 
I only needed to add a few net-new awaits, mostly in the instance where a new wrapper is created with some different prop value. 
In a file or two, i did deconstruct the nextTick function from vue vs calling vue.nextTick() etc but those were awaits were already there.

src/components/document-id/DocumentId.vue
```diff
@@ -0,0 +1,256 @@
...
+   * @param isValid - The validation value to emit.
+   */
+  @Emit('isValid')
+  emitIsValidDocumentId (isValid: boolean): boolean { return isValid }
+}
+</script>
+
+<style lang="scss" scoped>
+.title-label {
+  color: #212529FF;
```
severinbeauvais, line 243: I think we do have theme colours in this repo. (Also, the hex colours below.)

Also, this particular value looks wrong.

src/components/document-id/DocumentId.vue
```diff
@@ -0,0 +1,256 @@
...
+      switch (documentIdStatus) {
+        case 200:
+          this.docIdError = ['Must be unique number']
+          this.isVerifiedDocId = false
+          break
+        case 400:
+          this.docIdError = ['Document ID is invalid']
+          this.isVerifiedDocId = false
+          break
+        case 404:
```
severinbeauvais, line 190: We do have use a  library for HTTP response code elsewhere, if you want to use it.
-> http-status-codes

JazzarKarim (approved):
LGTM!

src/components/document-id/DocumentId.vue
cameron-eyds: Copy that, will double check!

src/components/document-id/DocumentId.vue
cameron-eyds: Will have a look!

src/components/document-id/DocumentId.vue
cameron-eyds: Fixed!

PR #266, opened 2025-01-30
Adding enums for draft status

TVWerdal:
*Issue #:*[ /bcgov/entity/23937](https://github.com/bcgov/entity/issues/23937)

Updating enums in entity-states.ts to include possible draft status values that appear during the lifecycle of a continue in filing.


src/enums/entity-states.ts
```diff
@@ -3,3 +3,15 @@ export enum EntityStates {
   HISTORICAL = 'HISTORICAL',
   LIQUIDATION = 'LIQUIDATION'
 }
```
severinbeauvais, line 5: Please update this enum as well. Its keys should match the new enum below.

severinbeauvais: No, wait.

This file should contain 2 enums: the state values, and the state strings.

Don't rename "EntityStates" since other UIs might already be importing and using it. Instead, I think you should update your UI to use "EntityStates" instead of "DraftStatus".

TVWerdal: Updated the enums file. Included the display strings used in BRD - these are currently included in a local enum in the BRD called business-state.ts. This will need to be changed through the BRD code so we are not duplicating the enum.

TVWerdal: Just thinking more on this - would we want "state display strings" to eventually go through i18n so they can be displayed in either Eng or Fr?

severinbeauvais: > Just thinking more on this - would we want "state display strings" to eventually go through i18n so they can be displayed in either Eng or Fr?

Yes, but that's a future problem :)

severinbeauvais: This code file looks as I expect. Resolving this conversation.

src/enums/package.json
severinbeauvais, line 1: Please un-commit this file.

We user a component versioning and dependency package called Lerna that manages the versions for us.

Once this PR is merged, I will run the Lerna steps to update the version numbers and publish your changes. (There are a few other admins in this repo in case I'm not available.)

severinbeauvais (approved):
Looks good. Merging.

severinbeauvais:
Enums v1.1.20 can now be imported.

The other updates are because THEY use the enums. Don't worry about them.

image

PR #268, opened 2025-02-04
Document Id: Integration Updates

cameron-eyds:
*Issue #:* /bcgov/entity#25408

*Description of changes:*
- Minor misc updates to Document Id as a result of integration work
- Updated styling
- added input character rule

### DIFF
--- a/src/components/document-id/DocumentId.vue
+++ b/src/components/document-id/DocumentId.vue
@@ -151,8 +151,8 @@ export default class DocumentId extends Vue {
 
     // Set up the headers for the API call
     const config = { headers: {
-        'Authorization': `Bearer ${sessionStorage.getItem(SessionStorageKeys.KeyCloakToken)}`,
-        'x-apikey': this.docApiKey }
+      'Authorization': `Bearer ${sessionStorage.getItem(SessionStorageKeys.KeyCloakToken)}`,
+      'x-apikey': this.docApiKey }
     }
 
     // Add the Account-Id header if the current account is available
--- a/src/components/document-id/DocumentId.vue
+++ b/src/components/document-id/DocumentId.vue
@@ -95,7 +95,7 @@
 import Vue from 'vue'
 import axios from 'axios'
 import { Component, Emit, Prop, Watch } from 'vue-property-decorator'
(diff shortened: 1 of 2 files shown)


doug-lovett (approved):
Looks great.

flutistar (approved):
LGTM

JazzarKarim (approved):
LGTM

src/components/document-id/DocumentId.vue
```diff
@@ -94,7 +95,7 @@
 import Vue from 'vue'
 import axios from 'axios'
 import { Component, Emit, Prop, Watch } from 'vue-property-decorator'
-import { SessionStorageKeys } from '@/enums/sbc-common-components-constants'
+import { SessionStorageKeys } from '@/enums'
```
severinbeauvais, line 98: Will this still work when this shared component is built into a UI? (Will this enum be present in `@/enums` in the UI?)

cameron-eyds: Yessir, will add the export to the enum index file.
The pattern seems to be consolidating the enums exports from the index, whether they are local/imported etc 
Seems best for reuse.

severinbeauvais: In Storybook (ie, in the current project), `@/enums` will find the new enum locally.

In, say, Create UI, `@/enums` will point to Create UI's enums folder -- but this enum isn't in that folder.

Option 1: change this to `from '@bcrs-shared-components/enums'.
Option 2: in Create UI (or wherever you need the enum), re-export the enum in enums/index.ts.

I prefer option 1, if it works with Storybook.

cameron-eyds: It didn't work previously but i did add it to the enums/index here as well. I can try option 1 with this change.

severinbeauvais: I believe some other shared components import shared enums and interfaces. Hopefully it works 🤞

cameron-eyds: Totally, it should now that sbc-constants were added to the bcrs enums/index!

cameron-eyds: Updated

PR #271, opened 2025-02-11
25522 fix_duplicate_address_info

kzdev420:
*Issue #:* /bcgov/entity#25522

*Description of changes:*

Fix - Address line 1 is duplicated in address line 2.

### DIFF
--- a/src/components/base-address/BaseAddress.vue
+++ b/src/components/base-address/BaseAddress.vue
@@ -83,6 +83,8 @@
         <div class="form__row">
           <v-textarea
             v-model="addressLocal.streetAddressAdditional"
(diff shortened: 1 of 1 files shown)


severinbeauvais:
The updated BaseAddress component has been published.

CompletingParty has also been updated and published. This happened automatically (via dependencies) because CompletingParty uses BaseAddress.

Therefore, please check the UIs for use of BaseAddress and CompletingParty and update their imports. NOTE: please check the changes between the in-use components and the latest ones to verify that there are no breaking changes. Most patch version updates (the last number in the version) are probably due to enum or interface updates, which generally are not breaking.

image

PR #272, opened 2025-02-11
25942 - added future effective state to entity states enum

JazzarKarim:
*Issue #:* /bcgov/entity#25942

*Description of changes:*
This is what things are going to look like in the table:
image


src/enums/entity-states.ts
```diff
@@ -4,6 +4,7 @@ export enum EntityStates {
   AWAITING_REVIEW = 'AWAITING_REVIEW',
   CHANGE_REQUESTED = 'CHANGE_REQUESTED',
   DRAFT = 'DRAFT',
+  FUTURE_EFFECTIVE = 'FUTURE_EFFECTIVE',
```
severinbeauvais, line 7: Is this an actual state (from Legal API), or is it more of a pseudo-state (computed by you)? If the latter then see if you can extend the enum right in the BRD code.

JazzarKarim: It's kinda more of the latter. Deal, going to leave this open for now and see if I can handle this there.

JazzarKarim:
Closing this PR as it's not needed. Ref: https://github.com/bcgov/business-ui/pull/67#discussion_r1953428552

PR #274, opened 2025-03-04
Document Record Service

flutistar:
*Issue #:* /bcgov/entity#25849

*Description of changes:*
Added enums and services for document service integration.


src/enums/document-types.ts
```diff
@@ -0,0 +1,40 @@
+export enum DocumentClassEnum {
+  CORP = 'CORP',
+  COOP = 'COOP',
+}
+
+export enum DocumetTypeEnum {
```
severinbeauvais, line 6: Typo.

src/enums/document-types.ts
```diff
@@ -0,0 +1,40 @@
+export enum DocumentClassEnum {
```
severinbeauvais, line 1: So far in this project, we've been naming enums like this: `DocumentClasses`.

Are you OK with this?

src/services/document-services.ts
```diff
@@ -0,0 +1,125 @@
+import Axios, { AxiosResponse } from 'axios'
```
severinbeauvais, line 1: I know the FeeServices imports Axios like this but I think it's bad practice.

Consider passing the existing Axios instance in as a prop, like this: https://github.com/bcgov/bcrs-shared-components/blob/e5ab4face748a4cfb6d16b1e566f6cabe88364b5/src/components/staff-comments/StaffComments.vue#L140

https://github.com/bcgov/business-filings-ui/blob/d61ffee682e1dd1b3a5931dc520e9fbb638530e4/src/components/EntityInfo/EntityMenu.vue#L6

src/services/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "@bcrs-shared-components/services",
-  "version": "1.0.35",
+  "version": "1.0.36",
   "publishConfig": {
     "access": "public"
   },
   "dependencies": {
-    "@bcrs-shared-components/enums": "^1.1.21",
+    "@bcrs-shared-components/enums": "^1.1.22",
```
severinbeauvais, line 8: As above, revert these changes. Lerna will do them automatically in step 14:
https://github.com/bcgov/bcrs-shared-components/blob/main/README.md

src/services/fee-services.ts
```diff
@@ -1,7 +1,6 @@
 import Axios from 'axios'
-import { Fee, FilingData, PayData } from '@/interfaces'
-import ConfigHelper from '@/utils/config-helper'
```
severinbeauvais, line 3: Why not use ConfigHelper?

src/services/package.json
severinbeauvais: To be clear: leave the extra dependency here but use current version numbers. Lerna will update the version numbers.

src/services/fee-services.ts
```diff
@@ -1,7 +1,6 @@
...
-import { SessionStorageKeys } from '@/enums/sbc-common-components-constants'
+import { Fee, FilingData, PayData } from '@bcrs-shared-components/interfaces'
+import { SessionStorageKeys } from '@bcrs-shared-components/enums/sbc-common-components-constants'
```
severinbeauvais, line 3: I _think_ this is correct, which makes me wonder how this worked previously.

What we want is:
- these imports work when we run this code in THIS project (eg, via unit tests or Storybook)
- these imports work when this component is imported into another project (ie, that project will have the right dependencies)

Please test these changes. If you cannot, I recommend not making these changes.

severinbeauvais (commented):
This will be nice to have as a shared component 👍 

Are there any possible unit tests that could be written (even though it's not a UI component)?

Is it possible to create a Storybook config for this (even though it's not a UI component)?

src/services/fee-services.ts
flutistar: Reasons for not using `configHelper`:

- There are no existing packages or functions that reference `configHelper`.
- We install packages independently (e.g., enums, services), and `utils` is unlikely to be installed as a child dependency.
- When testing @bcrs-shared-components/services in another project, an error occurred related to ConfigHelper.

src/services/fee-services.ts
flutistar: I wanted to avoid unnessaary changes but `fee-service` raises an error when I install bcrs-common/services in another project.

src/services/fee-services.ts
severinbeauvais: FeeServices and SbcFeeSummary (and others) were moved to this repo about 2 years ago in this PR: https://github.com/bcgov/bcrs-shared-components/pull/185.

I don't know if any UI imports these. You may be the first one.

I think you're right: `utils` is not installed as a dependency, because it isn't identified as a dependency (in package.json), because it's not being exported/published as its own package.

I suspect the FeeServices / SbcFeeSummary component was only tested in this repo -- where the files in `@/utils/config-helper` can be found. Or, if these were imported into another UI then that UI has the necessary files in its "utils" local folder.

It would be nice to get this working correctly without doing too much work (or repeating code, as you are doing here).

What do you think of moving the 2 utils files into this folder (since this is the only place they're used)? Then they will be part of the services package.

Do you think it would make more sense to export utils and make it a dependency here?

Do you have any other ideas?

src/services/fee-services.ts
severinbeauvais: What error, by the way?

src/services/fee-services.ts
severinbeauvais: PS - Here's the FeeSummary component, which uses the SbcFeeSummary component, which uses FeeServices. This is the local code working inside Storybook.

(Unfortunately the Pay API call is geting a 401, but otherwise I think it would work.)

This should KEEP WORKING after your changes.

flutistar: I agree with moving them into this folder.

severinbeauvais: Right on.

I forgot to attach the link to FeeSummary in Storybook: https://bcgov.github.io/bcrs-shared-components/?path=/story/component-feesummary--default

After your changes are complete, follow step 10 in the README to rebuild and run Storybook. You're looking for FeeSummary to keep working correctly (except for that 401).

src/services/fee-services.ts
severinbeauvais: Makes sense. The files can't be found because they're not in your local project.

I'll resolve this conversation and help you if you find anything else along the way. Looks good so far.

src/services/document-services.ts
cameron-eyds: The only catch i bumped into here is the interceptors from a local instance may not meet all requirements. 
ie missing account-id insertion into the header etc

Work around is to create another local instance but it might also make sense to make a service or component atomic so it doesn't need to rely on that local axios etc. 

Just considerations!

cameron-eyds (commented):
Looks great, outside the manual version bumps!

Read the readme good sir hahah 
Joking joking... sort of :)

JazzarKarim (approved):
Looks good to me (don't forget to revert the version changes back to their original state). Sev or I can publish the changes once the PR is merged (please wait for another approval before merging). Let us know 😄

src/services/document-services.ts
severinbeauvais: You can still provide the necessary headers in this file using a passed-in Axios instance, ie, this should still work fine:

image

As for interceptors, well if they're needed in the parent project then they're probably needed when this shared component is part of that project, yes?

cameron-eyds: For sure
Just calling out that passing in axios vs a local Axios instance with defined interceptors matters. 

If a service or component here has somewhat unique headers for it's requests, than the argument could be made to just make it atomic on this end, is all.

flutistar: The thing is, I might need to use these functions in any project that has doc-related components.
However, I believe we use useBcrosFetch in Nuxt applications(e.g. dashboard ui), and in that case, I need to create a new Axios instance to pass as a prop.

severinbeauvais: I'm debating whether the parent should manage any needed Axio instance, or whether the shared component should encapsulate that (though this feels like a "side effect").

Question: since you don't create an axios instance here, how does this work? I don't think `Axios` (TitleCase not lowercase) is an instance that you can use. Do you have this code working somewhere (eg, locally in a UI project)?

severinbeauvais: > If a service or component here has somewhat unique headers for it's requests,

Can you think of a real-life situation like this? It's a good case for a local Axios instance...

But do we need to attach a bearer token (JWT) to DRS calls?

flutistar: First of all, I tested these functions acroos create ui, edit ui, filings ui, and sbc-auth. 
So yes I updated those repos using these functions and like I mentioned while doing this I found the error on fee-service.

flutistar: image

I implemented this approach in `Create UI` to avoid adding the JWT token to the header by default. If I need to include a specific header for its requests, I prefer passing it as a parameter to a function and adding the header within the function itself.

That said, if using a local `axios` instance is the intended approach, I’m happy to adjust by passing axios as a function parameter as well.

severinbeauvais: Brandon, in your example above, did you use the current axios instance for the DRS service, and that's why you needed to override the interceptor?

Or were you using a separate axios instance in the DRS service? And you still needed to override the interceptor? And I still don't get how you could use axios in the service without first creating an instance.

Are we talking about separate scenarios?

severinbeauvais: I'm still not sure whether a local Axios instance (in your component) is desirable.

I do see the problem with the unwanted interceptor. Does it actually cause a problem?

Also, is there no auth for DRS calls?

severinbeauvais: Depending on how Axios is working currently with shared components, the local axios idea might be ideal or just as fraught with issues. I'm waiting for your response on my questions above to understand this and give an informed opinion. Thx.

flutistar: I need to override the intercepter because we get 401 response with JWT and we use apikey for authentication.

image

But yes I've included `axios` instance as a param for each function and I am writing unit test as requested.
Once the updated fucntions are fully test locally across create ui, edit ui, filings ui, and sbc-auth, i will make another commit and will inform you.

Thanks.

severinbeauvais: Are you saying that `import Axios, { AxiosResponse } from 'axios'` returns the existing instance (with interceptors), and that's why you don't need to "create" it but you do need to override the interceptor?

In a way, that's good, as it means we don't have to pass the axios instance into the component; the code above just returns it.

In a way, that's bad, because the existing instance can have interceptors, and it's really unfortunate you need to hack the UI's axios instance in order to use an external component (which, ideally, would be "atomic" as Cameron put it, or I would say "independent").

If the above is true then you don't need to pass in the prop, and let's keep this simple for now -- I will accept your existing code.

If I still have my lines crossed, let me know 😄

flutistar: Happy Friday Severin!

I just pushed a new commit which includes the below updates:
- passed axios instance as a param
- implemented unit test
- confirmed  fee-summary component is working on storybook
- tested on create-ui and edit-ui
- addressed additional issues.

This is the first time for me to work on npm package which is a great opportunity to learn new skills.
Please let me know if I missed anything.
Thank you for your help! :heart: :heart: :heart:

src/enums/sbc-common-components-constants.ts
```diff
@@ -26,7 +26,9 @@ export enum SessionStorageKeys {
   FasWebUrl = 'FAS_WEB_URL',
   AffidavitNeeded = 'AFFIDAVIT_NEEDED',
   GOVN_USER='AUTH_GOVN_USER',
-  PayApiUrl = 'PAY_API_URL'
+  PayApiUrl = 'PAY_API_URL',
+  DocApiUrl = 'DOC_API_URL',
+  DocApiKey = 'DOC_API_KEY'
```
severinbeauvais, line 31: _Next time_ this file is updated, we should alphabetize these keys to make them easier to look through!

flutistar: I will do it right now 😄

tests/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * 'Sleeps' for specified timeout. Must be called from async method.
+ * @param ms Delay to sleep, in milliseconds.
+ * @returns A promise to await upon.
+ */
+export function sleep (ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms))
+}
+
+export { instance as AxiosInstance }
```
severinbeauvais, line 19: Would you mind breaking this up into separate functions (named for their content)?

tests/utils.ts
```diff
@@ -0,0 +1,19 @@
+import axios from 'axios'
+
+const instance = axios.create()
+
+instance.interceptors.response.use(
+  (response) => response,
+  (error) => Promise.reject(error)
+)
```
severinbeauvais, line 8: Why do you need this for the tests?

src/enums/document-types.ts
```diff
@@ -0,0 +1,642 @@
...
+}
+
+export enum DocumentTypes {
+  DIRECTOR_AFFIDAVIT = 'DIRECTOR_AFFIDAVIT',
+  CORP_AFFIDAVIT = 'CORP_AFFIDAVIT',
+  COOP_MEMORANDUM = 'COOP_MEMORANDUM', // Memorandum
+  COOP_RULES = 'COOP_RULES', // Rules
+  COU = 'COU', // Court Order
+  COOP_MISC = 'COOP_MISC', // Correction Filing
+  COFI = 'COFI', // Cooperatives Miscellaneous Documents
```
severinbeauvais, line 20: Please alphabetize :)

severinbeauvais: (this whole list)

severinbeauvais: (within each section is OK)

You can also add whitespace between sections to make it easy to read.

src/services/document-services.ts
severinbeauvais, line 1: Just a thought -- instead of needing an interceptor in the main UI, why not add an interceptor here in this module?

https://github.com/axios/axios#multiple-interceptors

flutistar: I believe it should be better than what I did.
But what I can see instead of passing Axios instance as a param, moving axios utils function to shared-common-components repo is best practise.
I noticed that exactly same axios instance is used across at least 4 different repos but still sitting in those repos.

So I will move it to shared-common-components repo and will refactor all the related components.

Thanks

severinbeauvais: Wait. I want to chat with you about this before you start changing code.

src/services/utils/axios-instance.ts
```diff
@@ -0,0 +1,17 @@
...
+  request => {
+    request.headers['x-apikey'] = ConfigHelper.getFromSession('DOC_API_KEY')
+    request.headers['Account-Id'] = JSON.parse(ConfigHelper.getFromSession(SessionStorageKeys.CurrentAccount) || '{}')?.id || 0
+
+    return request
+  },
+  error => Promise.reject(error)
+)
+
+export { instance as axiosInstance }
```
severinbeauvais, line 17: This is OK but is only usable for the DRS service. It's OK for now.

src/services/document-services.ts
```diff
@@ -0,0 +1,140 @@
...
+import { SessionStorageKeys } from '@bcrs-shared-components/enums/sbc-common-components-constants'
+import ConfigHelper from './utils/config-helper'
+import { axiosInstance } from './utils'
+
+export default class DocumentService {
+  /**
+   * Retrieves the configuration for making an API request to the Document Record service.
+   * @returns an object containing the URL and headers for the DRS API request.
+   */
+  static requestConfig () {
```
severinbeauvais, line 11: Is this still needed?

src/services/document-services.ts
severinbeauvais: This has been figured out. Basically, this service will use its own Axios instance, with the interceptors it needs. It is a separate Axios instance from the one used in the parent UI.

severinbeauvais (approved):
Just a bit of cleanup, maybe.

@flutistar Let me know when I can merge and publish this for you.

flutistar:
Thanks, done.
@severinbeauvais

severinbeauvais:
There are a lot of component updates because the enums were updated and they depend on it. That's fine.

I updated the services package to 1.1.0 because you added a new one 🎉 

image

flutistar:
Thanks!

PR #276, opened 2025-03-06
26380 - Update BaseAddress Component

meawong:
*Issue #:* /bcgov/entity#26380 

*Description of changes:*
- Move Country selector above the Street Address field
- Set Canada as the default country
- Bind the Country selector to the Street Address field
- Hide the bar (i.e. AddressComplete logo, country selector) under the Street Address field

Demo in Filings-UI:
https://github.com/user-attachments/assets/c0c9735c-b91b-4d13-917a-8673f2cc9915

### DIFF
--- a/src/components/base-address/BaseAddress.vue
+++ b/src/components/base-address/BaseAddress.vue
@@ -369,6 +369,27 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
   }
 
   /**
+   * Watches changes to the Address Country and updates the schema accordingly.
+   */
+  @Watch('addressCountry')
+  onAddressCountryChanged (): void {
+    // skip this if component is called without a schema (eg, display mode)
+    if (this.schema) {
+      if (this.useCountryRegions(this.addressLocal['addressCountry'])) {
+        // we are using a region list for the current country so make region a required field
+        const addressRegion = { ...this.schema.addressRegion, required }
+        // re-assign the local schema because Vue does not detect property addition
+        this.schemaLocal = { ...this.schema, addressRegion }
+      } else {
+        // we are not using a region list for the current country so remove required property
+        const { required, ...addressRegion } = this.schema.addressRegion
+        // re-assign the local schema because Vue does not detect property deletion
+        this.schemaLocal = { ...this.schema, addressRegion }
+      }
+    }
+  }
+
+  /**
    * Watches changes to the Address Local object, to catch any changes to the fields within the address.
    * Will notify the parent object with the new address and whether or not the address is valid.
    */
--- a/src/components/base-address/BaseAddress.vue
+++ b/src/components/base-address/BaseAddress.vue
@@ -62,6 +62,26 @@
         lazy-validation
       >
         <div class="form__row">
+          <v-select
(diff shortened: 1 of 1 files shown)


src/components/base-address/BaseAddress.vue
```diff
@@ -203,7 +204,7 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
       addressCity: '',
       addressRegion: '',
       postalCode: '',
-      addressCountry: '',
+      addressCountry: 'CA',
```
severinbeauvais, line 207: Do you understand what this does exactly?

I ask because it looks like Canada is the default country even without this.

src/components/base-address/BaseAddress.vue
```diff
@@ -363,27 +368,6 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
-        // re-assign the local schema because Vue does not detect property addition
-        this.schemaLocal = { ...this.schema, addressRegion }
-      } else {
-        // we are not using a region list for the current country so remove required property
-        const { required, ...addressRegion } = this.schema.addressRegion
-        // re-assign the local schema because Vue does not detect property deletion
-        this.schemaLocal = { ...this.schema, addressRegion }
-      }
-    }
-  }
```
severinbeauvais, line 385: Don't we still need this?

src/components/base-address/BaseAddress.vue
meawong: This is on Dev, I don't see the default as Canada: 
image

severinbeauvais: Interesting.

If you search for a street address, the default is Canada. I don't know where that's set. Please see if there can be a single place that we set the default country.

I agree with you. The country should show Canada by default.

meawong: hmm I don't see Canada being set as the default anywhere else...I looked through the documentation and I thought it might be because of:
image
but even when I explicitly set this value to false, it defaults to Canada as the Country.

since we do want Canada to be the default is this ok? when we switch the country, the street address search populates results appropriately

severinbeauvais: Yes, this is OK. I just wanted to make sure we understand this (and didn't default the country in multiple places).

Please do one more test: edit an address that isn't in Canada, and see what the search defaults to.

meawong: Hopefully this video captures what you were asking to test:
https://github.com/user-attachments/assets/0153e023-9094-45f8-afd5-f0b01c625272

seeker25:
Cool, will be porting this to NuxtUI shortly hopefully. I'll let you know Sev. There's a few components we share and should work together on

src/components/base-address/BaseAddress.vue
severinbeauvais: Yes, great result! This shows that the default is Canada unless editing a non-null address.

Resolving.

severinbeauvais:
Megan, the new packages are published:

image

I will now try to update Storybook.

PR #311, opened 2026-05-07
33209 force international address auto complete in english char only

mruff-aeq:
#33209 : /bcgov/entity#33209

**Overview of Current Code State:**
The Canada Post AddressComplete library (addresscomplete-2.30.min.js) — vendored at public/js/addresscomplete-2.30.min.js (https://ws1.postescanada-canadapost.ca/js/addresscomplete-2.30.min.js), loaded via a global <script> tag in index.html. It runs entirely on window, attaches its API as window.pca, manages its own DOM (the suggestion dropdown is injected directly into document, not Vue-managed) and exposes its own event system via addressComplete.listen(eventName, callback).           

**Bug in `addresscomplete-2.30.min.js` that selects non-latin1 text:**
Unfortunately there is a bug in the `addresscomplete-2.30.min.js` in its' `populate` method by always selecting the **first** json object index, when working with the address auto complete for countries like China. When a Chinese address is searched by this endpoint http://ws1.postescanada-canadapost.ca/AddressComplete/Interactive/RetrieveFormatted/v2.10 , it returns an array of 2 json objects. 
Example:
<pre>
[{"Id":"CN|UP|B|N2808386","DomesticId":"","Language":"ENG","LanguageAlternatives":"ENG","Department":"","Company":"","SubBuilding":"","BuildingNumber":"","BuildingName":"","SecondaryStreet":"","Street":"","Block":"","Neighbourhood":"\u51E4\u5B89\u9547","District":"\u7D2B\u91D1\u53BF","City":"\u6CB3\u6E90\u5E02","Line1":"","Line2":"\u7D2B\u91D1\u53BF","Line3":"","Line4":"","Line5":"","AdminAreaName":"","AdminAreaCode":"","Province":"\u5E7F\u4E1C\u7701","ProvinceName":"\u5E7F\u4E1C\u7701","ProvinceCode":"","PostalCode":"517452","CountryName":"China","CountryIso2":"CN","CountryIso3":"CHN","CountryIsoNumber":"156","SortingNumber1":"","SortingNumber2":"","Barcode":"","POBoxNumber":"","Label":", \u7D2B\u91D1\u53BF\n517452 \u6CB3\u6E90\u5E02\nCHINA","Type":"","DataLevel":"Street","Field1":"","Field2":"","Field3":"","Field4":"","Field5":"","Field6":"","Field7":"","Field8":"","Field9":"","Field10":"","Field11":"","Field12":"","Field13":"","Field14":"","Field15":"","Field16":"","Field17":"","Field18":"","Field19":"","Field20":""},
{"Id":"CN|UP|B|N2808386","DomesticId":"","Language":"ENG","LanguageAlternatives":"ENG","Department":"","Company":"","SubBuilding":"","BuildingNumber":"","BuildingName":"","SecondaryStreet":"","Street":"","Block":"","Neighbourhood":"Feng An Zhen","District":"Zijin County","City":"Heyuan City","Line1":"","Line2":"Zijin County","Line3":"","Line4":"","Line5":"","AdminAreaName":"","AdminAreaCode":"","Province":"Guangdong Province","ProvinceName":"Guangdong Province","ProvinceCode":"","PostalCode":"517452","CountryName":"China","CountryIso2":"CN","CountryIso3":"CHN","CountryIsoNumber":"156","SortingNumber1":"","SortingNumber2":"","Barcode":"","POBoxNumber":"","Label":", Zijin County\n517452 HEYUAN CITY\nCHINA","Type":"","DataLevel":"Street","Field1":"","Field2":"","Field3":"","Field4":"","Field5":"","Field6":"","Field7":"","Field8":"","Field9":"","Field10":"","Field11":"","Field12":"","Field13":"","Field14":"","Field15":"","Field16":"","Field17":"","Field18":"","Field19":"","Field20":""}]
</pre>
As we can denote, the first json object contains Chinese characters in format`\u51E4\u5B89\u9547` , and the second json contains characters in latin-1 spec`Feng An Zhen`. 

Unfortunately, the current behavior in `addresscomplete-2.30.min.js`'s `populate` method  is to always pick the first item in the array, so the Chinese text will we passed to our form in this case.

**The Fix:**
In JavaScript, .test() is a method used with Regular Expressions to check if a string matches a specific pattern.

It returns a simple boolean:
 * true: The pattern was found in the string.
 * false: The pattern was NOT found.

 /^[\u0000-\u00ff]*$/.test(val)

 1. The Pattern: /^[\u0000-\u00ff]*$/ (The "Latin-1 only" rule).
 2. The Input: val (The text inside an address field, like "Victoria").
 3. The Action: It looks at val and asks: "Does this string consist ONLY of characters between \u0000 and \u00ff?"
 4. The Result:
     * If val is "Montreal", it returns true.
     * If val is "Méntrèal", it returns true (accents are in that range).
     * If val is "北京" (Beijing), it returns false (these characters are outside the range).

### DIFF
--- a/src/components/base-address/BaseAddress.vue
+++ b/src/components/base-address/BaseAddress.vue
@@ -509,8 +509,26 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
   /**
    * Callback to update the address data after the user chooses a suggested address.
    * @param address the data object returned by the AddressComplete Retrieve API
+   * UNDOCUMENTED! 2 args must be passed. address_ref contains first item in addresses, addresses contains entire return array.
    */
-  addressCompletePopulate (address: object): void {
+  addressCompletePopulate (addressRef: object, addresses: object[]): void {
+    // Select the first occurrence of the address only containing latin-1
+    const address = addresses.find(addr =>
+      // eslint-disable-next-line no-control-regex
+      Object.values(addr).every(val => typeof val !== 'string' || /^[\u0000-\u00ff]*$/.test(val))
+    ) || {
+      Line1: '',
+      Line2: '',
+      Line3: '',
+      Line4: '',
+      Line5: '',
+      City: '',
+      ProvinceCode: '',
+      ProvinceName: '',
+      PostalCode: '',
+      CountryIso2: ''
+    }
+
     const newAddressLocal: object = {}
 
     newAddressLocal['streetAddress'] = address['Line1'] || 'N/A'


mruff-aeq:
**Current: Address auto complete results in Chinese text being populated**
Screenshot 2026-05-07 at 4 03 57 PM
**My fix: Address auto complete results in English text being populated**
Screenshot 2026-05-07 at 4 01 11 PM

loneil (commented):
It's a lot of custom code (well, not really in the grand scheme of things) for specific character handling. Doesn't mean it's wrong to do it or anything, just that I don't know the address lookup component on these projects well enough yet.

Maybe @kialj876 does better (?) but would be worth knowing:
- Would other parts of the Registries ecosystem have this same problem (auth? new Nuxt views? Name Request?)
- If they don't, how did they setup the lookup to avoid that?

mruff-aeq:
> * Would other parts of the Registries ecosystem have this same problem (auth? new Nuxt views? Name Request?)
> * If they don't, how did they setup the lookup to avoid that?

It seems on other views, they disabled address lookup on countries other than Canada, but here it is the requirement. When dealing with Canadian addresses on other views, they would all go through the filter fine as the api returns latin-1 json for them.

src/components/base-address/BaseAddress.vue
```diff
@@ -506,11 +506,63 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
+   * Returns null when the picked record has no Id or window.addressCompleteKey
+   * is unset. Uses json3ex.ws — the CORS-enabled XHR endpoint the AddressComplete
+   * library itself uses — rather than json.ws, which is not CORS-enabled.
    */
-  addressCompletePopulate (address: object): void {
+  buildRetrieveFormattedUrl (picked: object): string | null {
+    const id = picked?.['Id']
+    const key = window['addressCompleteKey']
+    if (!id || !key) return null
+    return 'https://ws1.postescanada-canadapost.ca/AddressComplete/Interactive/RetrieveFormatted/v2.10/json3ex.ws' +
```
kialj876: I don't like that we have a harcoded url in here - I think we already get the info we need in the first call so I don't think this is necessary

src/components/base-address/BaseAddress.vue
```diff
@@ -543,6 +595,23 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
   }
 
+  /**
+   * Callback to update the address data after the user chooses a suggested address.
+   * If the library's picked record contains non-Latin-1 values (e.g. native-script
+   * Chinese), fetches the full RetrieveFormatted response and prefers a Latin-1
+   * variant when one exists. Falls back to the original picked record otherwise.
+   * @param address the data object returned by the AddressComplete Retrieve API
+   */
+  addressCompletePopulate (address: object): void {
```
kialj876: I think if you add a second arg here then it'll be the list of address items which will include the english version

src/components/base-address/BaseAddress.vue
```diff
@@ -543,6 +595,23 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
+    if (this.isLatin1Record(address)) {
+      this.applyAddressToForm(address)
+      return
+    }
+    this.fetchLatin1Alternate(address).then(latin1 => {
```
kialj876: instead of 'fetchLatin1Alternate' use the data you already have access to

kialj876 (changes requested):
Hey mark, nice work this was a bit of trickier one! 
I think you already have the data you need so there's no need for 'fetchLatin1Alternate' or 'buildRetrieveFormattedUrl' - let me know if you want to discuss

kialj876:
> It's a lot of custom code (well, not really in the grand scheme of things) for specific character handling. Doesn't mean it's wrong to do it or anything, just that I don't know the address lookup component on these projects well enough yet.
> 
> Maybe @kialj876 does better (?) but would be worth knowing:
> 
> * Would other parts of the Registries ecosystem have this same problem (auth? new Nuxt views? Name Request?)
> * If they don't, how did they setup the lookup to avoid that?

yeah @mruff-aeq can you ask vikas about creating a ticket to handle this similarly in our base layer repo - might be low priority but it currently just ignores the auto complete for non english countries there and I think this solution is better

src/components/base-address/BaseAddress.vue
```diff
@@ -509,8 +509,14 @@ export default class BaseAddress extends Mixins(ValidationMixin, CountriesProvin
...
    * Callback to update the address data after the user chooses a suggested address.
    * @param address the data object returned by the AddressComplete Retrieve API
+   * UNDOCUMENTED! 2 args must be passed. address_ref contains first item in addresses, addresses contains entire return array.
    */
-  addressCompletePopulate (address: object): void {
+   addressCompletePopulate (address_ref: object , addresses: object[]): void {
+    // Select the first occurrence of the address only containing latin-1
+    const address = addresses.find(addr =>
+      Object.values(addr).every(val => typeof val !== 'string' || /^[\u0000-\u00ff]*$/.test(val))
+    ) || address_ref
```
kialj876: I think this should probably set an address of empty strings if there isn't any latin-1 match instead of the address_ref? 
I think we want it to fall back to clearing any previous address info - would be an edge case that hopefully wouldn't happen, but we don't know for sure that canada post will always have an english option for every address

kialj876 (commented):
Looking good - please fix the lint and one minor comment below

kialj876 (approved):
Nice work this looks good to me now, thanks for the updates!

